PR #8, opened 2021-07-09
Use dynamic websocket host/port on dashboard

DaveWK:
This allows the dash to be accessed remotely behind a proxy


richardbrooks2000:
this seems to be a problem on firefox

PR #9, opened 2021-07-13
find dep libs

jhunsaker:


cmake/FindZSTD.cmake
```diff
@@ -0,0 +1,20 @@
+# taken from: https://github.com/facebook/folly/blob/master/CMake/FindZstd.cmake
+# should be apache 2.0, cf.: https://github.com/facebook/folly/blob/master/LICENSE
```
SEJeff, line 2: If you wouldn't mind, go to these urls and then hit `y`, which redirects you to the canonical commit. Otherwise, they can change things from underneath you and the comment will then be full of all lies.

richardbrooks2000:
this is not required

PR #24, opened 2021-08-06
replace subscribeSlot with getSlot

jhunsaker:


pc/rpc_client.hpp
```diff
@@ -366,6 +366,19 @@ namespace pc
       node_map_t nmap_;
     };
 
+    // get current slot
+    class get_slot : public rpc_request
+    {
+    public:
+      get_slot( commitment = e_finalized );
```
mass, line 373: Should we default to `e_confirmed`? That seems to be the most common/useful slot measurement

pc/manager.cpp
```diff
@@ -62,7 +62,8 @@ manager::manager()
   do_cap_( false ),
   do_tx_( true ),
   is_pub_( false ),
-  cmt_( commitment::e_confirmed )
+  cmt_( commitment::e_confirmed ),
+  sreq_{ { commitment::e_processed } }
```
mass, line 66: Why `processed` vs. `confirmed` here?

mass (approved):

pc/rpc_client.hpp
jhunsaker: since this is a simple C++ wrapper over the Solana RPC api, I think the C++ default should match the Solana default

pc/manager.cpp
jhunsaker: `slotSubscribe` is giving `processed` rather than `confirmed`. So this is giving the same behavior. Open to changing to `confirmed`, but I feel that would be a separate change. In practice, I noticed these tend to be 1 apart.

PR #32, opened 2021-08-13
pyth.cpp: upd_price takes price value, confidence and symbol status

drozdziak1:
This change allows writing `./pyth upd_price_val` which takes an explicit price, confidence interval and status value for publishing.

### DIFF
--- a/pcapps/pyth.cpp
+++ b/pcapps/pyth.cpp
@@ -49,6 +49,8 @@ int usage()
   std::cerr << "  upd_product      <product.json> [options]" << std::endl;
   std::cerr << "  upd_price        <price_key> [options]"
             << std::endl;
+  std::cerr << "  upd_price_val    <price_key> <value> <confidence> <status> [options]"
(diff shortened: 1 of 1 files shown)


pcapps/pyth.cpp
```diff
@@ -801,14 +801,26 @@ int on_upd_test( int argc, char **argv )
 
 int on_upd_price( int argc, char **argv )
 {
-  if ( argc < 2 ) {
+  if ( argc < 5 ) {
```
mass, line 804: Could this change be made backwards compatible? e.g.

If argc == 2, then `ptr->update()`, else if argc == 5, then `ptr->update(px, ci, st)`, else err

drozdziak1: Moved to a new subcommand `upd_price_val`

mass (commented):
Looks great, thank you! Just a couple of comments:

mass (approved):

PR #37, opened 2021-08-18
poll program accounts

jhunsaker:
this is a partial change - the code to activate the polling is not included yet (but was tested locally)

this refactor should not change existing behavior, but allows for code deduplication

### DIFF
--- a/pc/rpc_client.hpp
+++ b/pc/rpc_client.hpp
@@ -434,13 +434,15 @@ namespace pc
       uint64_t    slot_;
     };
 
-    // account data subscription
-    class account_subscribe : public rpc_subscription
+    // base class for account updates
+    class account_update : public rpc_subscription
     {
     public:
-      // parameters
+      account_update();
+
+      // parameters / results
+      pub_key const* get_account() const;
       void set_account( pub_key * );
-      void set_commitment( commitment );
 
       // results
       uint64_t get_slot() const;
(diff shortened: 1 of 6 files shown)


jhunsaker:
additional follow up changes are to add filtering - this is grabbing all program accounts - including larger accounts that don't need to be grabbed as frequently

pc/rpc_client.hpp
```diff
@@ -474,61 +470,64 @@ namespace pc
...
+      // parameters
+      void set_commitment( commitment );
+
+      account_subscribe();
+      void request( json_wtr& ) override;
+      void response( const jtree& ) override;
+      bool notify( const jtree& ) override;
+
+    private:
+      commitment  cmt_;
```
mass, line 493: Should `cmt_` be moved into the base class seeing all the derived classes have it?

pc/rpc_client.hpp
```diff
@@ -450,22 +452,16 @@ namespace pc
...
 
-      account_subscribe();
-      void request( json_wtr& ) override;
-      void response( const jtree& ) override;
-      bool notify( const jtree& ) override;
-
-    private:
-      pub_key    *acc_;
+    protected:
+      pub_key     acc_;
```
mass, line 456: Why did this change from a `pub_key*` to a `pub_key`?

mass (approved):

pc/rpc_client.hpp
jhunsaker: yes, thanks, I noticed that but didn't want to repeat the testing for a small change - will include in the next round

pc/rpc_client.hpp
jhunsaker: because it's an output of the program calls whereas in the account_subscribe it's a parameter

PR #66, opened 2021-09-26
add status to test results

jhunsaker:


jhunsaker:
also, additional tests from @jbezaire

pctest/test_qset.cpp
```diff
@@ -146,12 +146,29 @@ int main( int argc,char** argv )
   }
   upd_aggregate( px, slot+1 );
 
+  char const* status = "invalid value";
+  switch ( px->agg_.status_ ) {
```
mass, line 150: Could `symbol_status_to_str` be used for this?

mass (approved):

pctest/test_qset.cpp
jhunsaker: thanks, missed that function (was grepping for the constants) let's move that out of `rpc_client` and make it a bit more robust then we can use it more places

PR #67, opened 2021-09-27
filter out quoters far from median

jhunsaker:
with @jbezaire

### DIFF
--- a/program/src/oracle/sort.c
+++ b/program/src/oracle/sort.c
@@ -1,0 +1,44 @@
+#include "sort.h"
+
+static inline void swap_int64( int64_t *a, int64_t *b )
+{
+  int64_t const temp = *a;
+  *a = *b;
+  *b = temp;
+}
+
+static inline int partition_int64( int64_t* v, int i, int j )
+{
+  int const p = i;
+  int64_t const pv = v[ p ];
+
+  while ( i < j ) {
+    while ( v[ i ] <= pv && i <= j ) {
+      ++i;
+    }
+    while ( v[ j ] > pv ) {
+      --j;
+    }
+    if ( i < j ) {
+      swap_int64( &v[ i ], &v[ j ] );
+    }
+  }
+  swap_int64( &v[ p ], &v[ j ] );
+  return j;
+}
+
+static void qsort_help_int64( int64_t* const v, int i, int j )
+{
+  if ( i >= j ) {
+    return;
+  }
+  int p = partition_int64( v, i, j );
+  qsort_help_int64( v, i, p - 1 );
+  qsort_help_int64( v, p + 1, j );
+}
+
+void qsort_int64( int64_t* const v, unsigned const size )
+{
+  qsort_help_int64( v, 0, ( int )size - 1 );
+}
+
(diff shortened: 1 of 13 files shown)


jbezaire (dismissed):

program/src/oracle/sort.c
```diff
@@ -0,0 +1,44 @@
...
+  int const p = i;
+  int64_t const pv = v[ p ];
+
+  while ( i < j ) {
+    while ( v[ i ] <= pv ) {
+      ++i;
+    }
+    while ( v[ j ] > pv ) {
+      --j;
+    }
```
mass, line 21: Are these guaranteed not to slide past `[0,size)`?

program/src/oracle/upd_aggregate.h
```diff
@@ -294,31 +298,59 @@ static void upd_aggregate( pc_price_t *ptr, uint64_t slot )
...
+  uint32_t nprcs = 0;
+  int64_t prcs[ PC_COMP_SIZE * 2 ];
+  for ( uint32_t i = 0; i < numv; ++i ) {
+    pc_price_comp_t const *iptr = &ptr->comp_[ vidx[ i ] ];
+    int64_t const price = iptr->agg_.price_;
+    int64_t const conf = ( int64_t )( iptr->agg_.conf_ );
+    prcs[ nprcs++ ] = price - conf;
+    prcs[ nprcs++ ] = price + conf;
+  }
+  qsort_int64( prcs, nprcs );
```
mass, line 327: Maybe we should insert all the `price-conf` at 0 + i and `price+conf` at `numv + i`, such that `prcs` (maybe) requires less instructions during sorting?

program/src/oracle/upd_aggregate.h
```diff
@@ -294,31 +298,59 @@ static void upd_aggregate( pc_price_t *ptr, uint64_t slot )
...
+
+  uint32_t numa = 0;
+  uint32_t aidx[PC_COMP_SIZE];
+
+  if ( nprcs ) {
+    int64_t const mprc = ( prcs[ numv - 1 ] + prcs[ numv ] ) / 2;
```
mass, line 333: If I am understanding correctly, we only need the median two values from the sorted `prcs` array. I think we could just quick-select for those rather than sorting the whole array. But, not sure how much we need to optimize for instructions at this place in the code

mass (approved):

program/src/oracle/sort.c
jhunsaker: the first loop is not, added a condition
the second loop is, because `pv = v[ i ]`

program/src/oracle/upd_aggregate.h
jhunsaker: this was considered, but with the current sort, the pivot is the first element

program/src/oracle/upd_aggregate.h
jhunsaker: right, this was discussed, but qsort as library func (and subsequent copy/paste here) was convenient
we will revisit the sort after issue #64

mass (approved):

jbezaire (approved):

PR #72, opened 2021-10-02
proper lprice/uprice sorting in confidence calculation

jhunsaker:


jhunsaker:
with @jbezaire

program/src/oracle/upd_aggregate.h
```diff
@@ -313,38 +325,58 @@ static inline void upd_aggregate( pc_price_t *ptr, uint64_t slot )
...
 
   // compute diff in weighted median between upper and lower price bounds
-  wgt_ptile( uprice, qs->uprice_, ptile, qs );
-  wgt_ptile( lprice, qs->lprice_, ptile, qs );
+  pd_t prices[ PC_COMP_SIZE ];
+  pd_t weights[ PC_COMP_SIZE ];
+  // sort upper prices and weights
+  for ( uint32_t i = 0; i < numa; ++i ) {
+    uint32_t j = i;
+    for ( ; j > 0 && pd_lt( &qs->uprice_[ i ], &prices[ j - 1 ], qs->fact_ ); --j ) {
```
mass, line 350: I wonder if using the qsort you added would be more efficient w.r.t instructions (add all, then qsort once)

mass (approved):

program/src/oracle/upd_aggregate.h
jhunsaker: Thanks - started down that route but it added complexity (we have to move weight values in parallel), and the `uprice_` is already mostly sorted, so decided to use insertion sort for now.

PR #88, opened 2021-10-12
Basic template for fuzzing

jayantk:
This PR adds a template for property-based testing the solana program using AFL. The core idea is that we make a command-line program for each property we want to test (e.g., the sum of two positive numbers is positive), then we build that program with AFL instrumentation, and run it with AFL fuzzer. I built everything into a docker container so it's easy to reproduce the whole setup.

I also drive-by added a dockerfile for building the program on M1 macs, because there's no other way to do it.

Note that this approach to fuzzing tests the C code compiled for x86. I think we also need some fuzz tests for BPF running on the solana validator, but I will set that up later. The analogy should be that the current thing is unit tests for the code, and the BPF thing will be integration tests for Pyth.

I'm not sure this is the best way to organize the repo, so definitely open to suggestions. I also don't write a lot of C++ or dockerfiles, so please let me know if I'm doing anything wrong.


CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ set( CMAKE_INCLUDE_CURRENT_DIR ON )
 include_directories( program/src/ )
 
 # gcc compiler/linker flags
-add_compile_options( -ggdb -Wall -Wextra -Werror -m64 )
+# FIXME: add -Werror back
+add_compile_options( -ggdb -Wall -Wextra -m64 )
```
tony-ricciardi, line 19: This should no longer be necessary once we merge https://github.com/pyth-network/pyth-client/pull/89.

PR #101, opened 2021-10-19
enable signed conversion warning

jhunsaker:


jhunsaker:
this pull request will be followed by another request that converts most or all of these debug assertions to actual runtime assertions in the release build

PR #110, opened 2021-11-02
Update docs for m1 mac build

jayantk:


README.md
```diff
@@ -17,14 +17,30 @@ apt install libzstd-dev
...
+```
+
+If you are using an M1 mac, you will need to build this repository inside a linux docker container, as the build depends on
+some linux-specific libraries. You can run the following command to open a shell in a linux docker container with the current
+directory mounted:
 
-# run unit tests
-ctest
 ```
+export PYTH_REPO=`pwd`
```
tony-ricciardi, line 28: `PYTH_REPO=/path/to/host/pyth-client` might be more clear. If you'd rather to stick with `pwd`, it should be `"$( pwd )"` to handle whitespace and be consistent with the `podman` command below.

README.md
```diff
@@ -17,14 +17,30 @@ apt install libzstd-dev
...
-mkdir build
-cd build
-cmake ..
-make
+./scripts/build.sh
+```
+
+If you are using an M1 mac, you will need to build this repository inside a linux docker container, as the build depends on
+some linux-specific libraries. You can run the following command to open a shell in a linux docker container with the current
+directory mounted:
```
tony-ricciardi, line 25: I would remove the first sentence, since this isn't specific to OS X. I use a similar `podman` command on RHEL, and more generally users should always be building & running BPF binaries inside a docker environment.

README.md
```diff
@@ -17,14 +17,30 @@ apt install libzstd-dev
...
+  --mount "type=bind,src=${PYTH_REPO},target=/home/pyth/pyth-client" \
+  --userns=host \
+  --user="$( id -ur ):$( id -gr )" \
+  --platform linux/amd64 \
+  $IMAGE \
+  /bin/bash -l
+```
+
+This command runs a recent pyth-client docker image that already has the necessary dependencies installed.
+Therefore, once the container is running, all you have to do is run `cd pyth-client && ./scripts/build.sh`.
```
tony-ricciardi, line 43: Also worth noting that the `--user` argument causes changes to source files, build directories, etc., within the mounted `~/pyth-client/` to be persisted to the host (generally desired).

tony-ricciardi (dismissed):

Issue #113, opened 2021-11-05
Pyth client websocket interface should lower case headers before comparing them


linuskendall:
Headers in HTTP should be processed case-insentively, however the pythd proxy relies on the headers being capitalised. This causes issues when attempting to place the Pythd websocket interface behind a proxy (or using a client that uses lower-case headers).  

Relevant locations in the pythd source (AFAICT):
```
./net_socket.cpp:1059:  if ( !get_header_val( "Sec-WebSocket-Key", key ) ) {
./net_socket.cpp:1028:          0 == __builtin_strncmp( "Upgrade", hdr, hlen ) &&
```

The first causes a 'hang' (the connection just stalls) if Sec-WebSocket-Key is specified with any other capitalisation. The second case causes websocket upgrades to not be processed at all and the pythd API returns 404 (as it doesn't attempt websocket upgrades).

PR #116, opened 2021-11-07
http headers are case insensitive

jhunsaker:


pc/net_socket.cpp
```diff
@@ -1029,16 +1031,19 @@ bool http_server::parse( const char *ptr, size_t len, size_t& res )
     if ( !find( CR, ptr, end ) )  return false;
     assert( hdr_end >= hdr );
     const size_t hlen = static_cast< size_t >( hdr_end - hdr );
+    std::string hdr_str{ hdr, hlen };
```
tony-ricciardi, line 1034: `const`

pc/net_socket.cpp
```diff
@@ -1029,16 +1031,19 @@ bool http_server::parse( const char *ptr, size_t len, size_t& res )
...
+    for ( char& c : hdr_str ) {
+      c = std::toupper( static_cast< unsigned char>( c ) );
```
tony-ricciardi, line 1036: whitespace - should be `< unsigned char >`

tony-ricciardi (approved):

pc/net_socket.cpp
jhunsaker: it's modified

tony-ricciardi (approved):

jhunsaker:
this is a partial fix for issue #113 - more header logic should be made case-insensitive

PR #119, opened 2021-11-12
add bisontrails example nginx config

jhunsaker:


doc/example-nginx-bisontrails.conf
```diff
@@ -0,0 +1,43 @@
+# generate YOUR_BASE64_USERPASS with `echo -n user:password | base64`
```
SEJeff, line 1:
```suggestion
# generate YOUR_BASE64_USERPASS with `echo -n user:password | base64 -w0`
```

Probably want to do this just in case. It will wrap on occasion and I already asked BT to update their docs on this when my first one wrapped.

SEJeff (approved):
Nice stuff. Only one minor nit sir.

SEJeff (approved):

PR #127, opened 2021-12-03
test price account

jhunsaker:


pyth/tests/conftest.py
```diff
@@ -228,3 +230,63 @@ def pyth_add_product(solana_test_validator, pyth_dir, pyth_init_mapping):
...
+    with fdopen(fd, 'w') as f:
+        json.dump([ltc], f)
+    cmd = [
+        'pyth_admin', 'upd_product', path,
+        '-r', 'localhost',
+        '-k', pyth_dir,
+        '-c', 'finalized',
+    ]
+    check_call(cmd)
+    os.remove(path)
```
tony-ricciardi, line 259: This won't remove the temp file on an exception. I think it should be:
```
fd, path = mkstemp(suffix='.json', prefix='ltc_')
try:
    with fdopen(fd, 'w') as f:
        json.dump([ltc], f)
    cmd = [
        'pyth_admin', 'upd_product', path,
        '-r', 'localhost',
        '-k', pyth_dir,
        '-c', 'finalized',
    ]
    check_call(cmd)
finally:
    os.remove(path)
```

jhunsaker: that was intentional - if it fails we want to see the input

tony-ricciardi: Ok, but logging the file contents on exception before removing would probably be less annoying than remembering to manually remove the file after each failure.

tony-ricciardi (approved):

PR #133, opened 2021-12-29
add batching support (incremental commit)

jhunsaker:


pc/rpc_client.cpp
```diff
@@ -879,20 +879,104 @@ void rpc::upd_price::build_tx( bincode& tx )
...
+}
+
 void rpc::upd_price::build( net_wtr& wtr )
 {
-  bincode tx;
-  ((tx_wtr&)wtr).init( tx );
-  build_tx( tx );
-  ((tx_wtr&)wtr).commit( tx );
+  upd_price* upds[] = { this };
+  build( wtr, upds, 1 );
```
dragonfrond, line 947: Worth checking the return value of `build` here and at least logging if it fails

dragonfrond (approved):

pc/rpc_client.cpp
jhunsaker: it can't fail in this case, but I agree we can add a check

PR #134, opened 2021-12-30
support batch price updates

jhunsaker:


dragonfrond (approved):
One change requested, but otherwise fine

PR #140, opened 2022-01-24
Batch updates in update_price JRPC call

tompntn:
This PR changes the semantics of the `update_price` pythd JRPC call. Previously, updates were immediately sent to Solana (or `pyth_tx`) using a new transaction for each update. Instead, we buffer all received updates inside pythd, and send all buffered updates in bulk whenever a new slot is available. This uses less transactions at a cost of slightly increased latency.

I have tried to follow existing code styles and conventions as much as possible.

A follow-up PR will modify the on-chain program, adding a batch update endpoint which will only fail the transaction if all the individual price updates in that batch fail. This will ensure publishing failures on one product won't block updates for unaffected products.

### DIFF
--- a/pc/manager.cpp
+++ b/pc/manager.cpp
@@ -746,6 +746,12 @@ void manager::on_response( rpc::get_slot *res )
   if (
     has_status( PC_PYTH_RPC_CONNECTED )
   ) {
+    
+      // New slot received, so flush all pending updates for all active users
+      for( user *uptr = olist_.first(); uptr; uptr = uptr->get_next() ) {
+        uptr->send_pending_upds();
+      }
+
     if ( sub_ ) {
       sub_->on_slot_publish( this );
     }
--- a/pc/request.cpp
+++ b/pc/request.cpp
@@ -708,12 +708,27 @@ bool price::send( price *prices[], const unsigned n )
   for ( unsigned i = 0, j = 0; i < n; ++i ) {
     price *const p = prices[ i ];
     if ( PC_UNLIKELY( ! p->init_ && ! p->init_publish() ) ) {
+      PC_LOG_ERR( "failed to initialize publisher" )
+        .add( "price_account", *p->get_account() )
+        .add( "product_account", *p->prod_->get_account() )
+        .add( "symbol", p->get_symbol() )
+        .add( "price_type", price_type_to_str( p->get_price_type() ) ).end();
       continue;
     }
     if ( PC_UNLIKELY( ! p->has_publisher() ) ) {
+      PC_LOG_ERR( "missing publish permission" )
+        .add( "price_account", *p->get_account() )
+        .add( "product_account", *p->prod_->get_account() )
+        .add( "symbol", p->get_symbol() )
+        .add( "price_type", price_type_to_str( p->get_price_type() ) ).end();
       continue;
     }
     if ( PC_UNLIKELY( ! p->get_is_ready_publish() ) ) {
+      PC_LOG_ERR( "not ready to publish - check rpc / pyth_tx connection" )
+        .add( "price_account", *p->get_account() )
+        .add( "product_account", *p->prod_->get_account() )
+        .add( "symbol", p->get_symbol() )
+        .add( "price_type", price_type_to_str( p->get_price_type() ) ).end();
       continue;
     }
     manager *const mgr = p->get_manager();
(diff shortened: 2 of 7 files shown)


pc/user.cpp
```diff
@@ -204,22 +206,18 @@ void user::parse_upd_price( uint32_t tok, uint32_t itok )
...
-    } else {
-      add_error( itok, PC_JSON_INVALID_REQUEST, "unknown error" );
+    // Add the updated price to the pending updates
+    sptr->update_no_send( price, conf, stype, false );
+    if(std::find(pending_vec_.begin(), pending_vec_.end(), sptr) == pending_vec_.end()) {
+      pending_vec_.emplace_back( sptr );
     }
+
+    // Send the result back
+    //  TODO: should we send this result? As the semantics have now changed.
```
jayantk, line 216: this seems fine -- I think we don't really have a choice, because not sending this result would change the API.

jayantk (commented):
Seems reasonable to me. My C++ is terrible though, so would be good to let david look as well.

Main question is what's our testing strategy?

pc/manager.cpp
```diff
@@ -746,6 +746,14 @@ void manager::on_response( rpc::get_slot *res )
   if (
     has_status( PC_PYTH_RPC_CONNECTED )
   ) {
+    
+      // New slot received, so flush all pending updates for all active users
+      for( user *uptr = olist_.first(); uptr; ) {
+        user *nptr = uptr->get_next();
+        uptr->send_pending_upds();
+        uptr = nptr;
```
jhunsaker, line 754: why is the loop logic not part of the for loop?

pc/request.cpp
```diff
@@ -708,12 +708,15 @@ bool price::send( price *prices[], const unsigned n )
   for ( unsigned i = 0, j = 0; i < n; ++i ) {
     price *const p = prices[ i ];
     if ( PC_UNLIKELY( ! p->init_ && ! p->init_publish() ) ) {
+      PC_LOG_ERR( "failed to initialize publisher" );
```
jhunsaker, line 711: do these conditions occur naturally, like on startup? might want to make these WARN if so?

pc/user.cpp
```diff
@@ -333,6 +331,21 @@ void user::parse_get_product( uint32_t tok, uint32_t itok )
   add_tail( itok );
 }
 
+void user::send_pending_upds()
+{
+  if ( pending_vec_.empty() ) {
+    return;
+  }
+
+  if ( price::send( &pending_vec_[ 0 ], pending_vec_.size()) == false ) {
```
jhunsaker, line 340: ! price:send...

pc/manager.cpp
jhunsaker: another impl option is to have user receive the new slot callback? what's reasoning behind this impl decision?

jayantk: We want to preserve the existing API so that we can quickly roll these changes out to publishers.

tompntn: That iteration pattern was used elsewhere in the codebase so I adopted it for consistency - but yes it could definitely be made simpler.

pc/request.cpp
tompntn: They were previously logged at error level, and I think it's safer to keep it that way. Having a few spurious errors at statup is better than missing important warnings if something breaks after `pythd` has been running for a while. Open to other opinions though 😄

pyth/tests/test_update_price.py
```diff
@@ -0,0 +1,105 @@
...
+                'trading')
+
+        time.sleep(60)
+
+        # Check that the price has been updated
+        for product in PRODUCTS.keys():
+            product_acc = await get_product(pyth_init_product[product])
+            publisher_acc = get_publisher_acc(product_acc)
+
+            assert publisher_acc['price'] == new_values[product]['price']
```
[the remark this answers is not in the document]
jayantk: I think it would be good to test the end-to-end API somewhere. Let's use pythd exactly as a publisher would: register for the price callbacks and then submit updates in response. This is probably not easy to do in a unit test, but could be a mock publisher or something. 

(That doesn't need to happen as part of this PR.)

jayantk (dismissed):
LGTM!

pyth/tests/test_update_price.py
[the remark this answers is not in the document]
tompntn: > why is this the value in new_values and not 1?

The second update is just to crank the prices.

pc/user.cpp
```diff
@@ -333,6 +330,21 @@ void user::parse_get_product( uint32_t tok, uint32_t itok )
   add_tail( itok );
 }
 
+void user::send_pending_upds()
+{
+  if ( pending_vec_.empty() ) {
+    return;
+  }
+
+  if ( !price::send( pending_vec_.data(), pending_vec_.size()) ) {
```
Reisen, line 339: 👌

Reisen (approved):

jayantk (approved):

PR #141, opened 2022-01-26
Accessors for base/quote currencies

jayantk:
These accessors make it easy for people to avoid using the symbol

### DIFF
--- a/pc/request.cpp
+++ b/pc/request.cpp
@@ -505,6 +505,16 @@ str price::get_symbol()
   return prod_->get_symbol();
 }
 
+str price::get_base()
+{
+  return prod_->get_base();
+}
+
+str price::get_quote()
+{
+  return prod_->get_quote();
+}
+
 bool price::get_attr( attr_id aid, str& val ) const
 {
   return prod_->get_attr( aid, val );
--- a/pc/request.cpp
+++ b/pc/request.cpp
@@ -295,6 +295,20 @@ str product::get_symbol()
   return sym;
 }
 
+str product::get_base()
+{
+  str base;
+  get_attr( attr_id( "base" ), base);
+  return base;
+}
+
+str product::get_quote()
+{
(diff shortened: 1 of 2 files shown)


viniiciusconceicao (approved):

pc/request.cpp
```diff
@@ -295,6 +295,20 @@ str product::get_symbol()
...
 }
 
+str product::get_base()
+{
+  str base;
+  get_attr( attr_id( "base" ), base);
+  return base;
+}
+
+str product::get_quote()
```
dragonfrond, line 305: Should really be named `get_quote_currency` since `get_quote` implies something entirely different

pc/request.hpp
```diff
@@ -243,6 +247,12 @@ namespace pc
     // convenience wrapper equiv to: get_product()->get_symbol()
     str get_symbol();
 
+    // Get the base currency for the product (from the list of product attributes).
+    str get_base();
```
dragonfrond, line 251: This is not the base currency but the base asset, rename `get_base_asset` (might not be a currency)

pc/request.cpp
```diff
@@ -295,6 +295,20 @@ str product::get_symbol()
...
+{
+  str quote;
+  get_attr( attr_id( "quote_currency" ), quote);
```
dragonfrond, line 308: Does this fail gracefully if the attribute does not exist?

dragonfrond (approved):

pc/request.cpp
jayantk: I'm copy pasting `get_symbol` here. 

That said, my reading of the code is that it does fail gracefully. If the attribute is not found, `attr_id` returns the id of 0 (which, confusingly, represents an invalid attribute id, not the first attribute). get_attr on that id should give you a string whose char* is null.

PR #139, opened 2022-01-12
Merge new aggregation logic into pyth-client (on behalf of kbowers)

jayantk:

PR #145, opened 2022-03-03
Remove old aggregation logic.

Reisen:
This removes the old aggregation logic and simply stores the price and confidence directly in the aggregate. This follows up on the aggregation logic in v2 extracted from #139 and removes the grafting on input to the TWAP calculation.

### DIFF
--- a/program/src/oracle/upd_aggregate.h
+++ b/program/src/oracle/upd_aggregate.h
@@ -149,42 +149,6 @@ static inline void upd_twap(
   upd_ema( &ptr->twac_, conf, conf, nslots, qs, ptr );
 }
 
-// compute weighted percentile
-static void wgt_ptile(
-  pd_t * const res
-  , const pd_t * const prices, const pd_t * const weights, const uint32_t num
-  , const pd_t * const ptile, pc_qset_t *qs )
-{
-  pd_t cumwgt[ PC_COMP_SIZE ];
-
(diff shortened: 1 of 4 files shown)


program/src/oracle/upd_aggregate.h
```diff
@@ -266,180 +229,13 @@ static inline void upd_aggregate( pc_price_t *ptr, uint64_t slot )
...
-          aidx[ j ] = aidx[ j - 1 ];
-        }
-        aidx[ j ] = idx;
-      }
-    }
-  }
-
-  // zero quoters
-  ptr->num_qt_ = numa;
-  if ( numa == 0 || numa < ptr->min_pub_ || numa * 2 <= numv ) {
```
jayantk, line 298: regarding the test changes, it looks to me like the tests are running with `min_pub_ == 0`. The reason the tests previously gave a result of `status=Unknown` is because of this `numa * 2 <= numv` check, which shouldn't be necessary with the new aggregation algorithm. (The old algorithm ran a prefilter on the active publishers to only include the ones whose price was close to the median price. This check is verifying that the prefilter didn't throw out too many publishers.)

Can you double-check that the tests are running with `min_pub_ == 0`? If they are, then the test changes are ok.

jayantk (approved):
lgtm. I think the test changes are fine, but do double check per the comment below.

PR #146, opened 2022-03-03
Update max send latency

jayantk:
We have some symbols that are slightly flaky right now. They go down for ~1 second at a time then come back up. This change lets us tolerate prices that are slightly older, which should reduce flakiness.


jayantk:
turns out this doesn't help uptime all that much.

PR #149, opened 2022-03-22
Add Compiler flag to ignore deprecated OpenSSL declarations

DaveWK:
When compiling against OpenSSL versions => 3.0, the build will error out
due to the usage of methods like SHA256_Update, SHA256_Init, etc..

See: https://wiki.openssl.org/index.php/OpenSSL_3.0

While these functions should probably be ported to their new
equivalents, this at least allows for the client to be compiled.


jayantk (approved):
thanks!

PR #152, opened 2022-04-05
Store value of previous price aggregate if the status is trading

tompntn:
This PR changes the semantics of the `pc_price.prev_slot_`, `pc_price.prev_price_` and `pc_price.prev_conf_` fields to only store the value of the previous price aggregate if the previous price aggregate had a status of TRADING. We also store a new piece of data, the Solana unix timestamp of this previous aggregate, for convenience.

### DIFF
--- a/program/src/oracle/test_oracle.c
+++ b/program/src/oracle/test_oracle.c
@@ -569,6 +569,9 @@ Test( oracle, upd_aggregate ) {
   cr_assert( px->twac_.val_ == 10 );
   cr_assert( px->num_qt_ == 1 );
   cr_assert( px->timestamp_ == 1 );
+  cr_assert( px->prev_slot_ == 0 );
+  cr_assert( px->prev_price_ == 0 );
+  cr_assert( px->prev_conf_ == 0 );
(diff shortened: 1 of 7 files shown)


program/src/oracle/test_oracle.c
```diff
@@ -604,24 +619,37 @@ Test( oracle, upd_aggregate ) {
...
 
   // check what happens when nothing publishes for a while
-  upd_aggregate( px, 1026 );
+  upd_aggregate( px, 1026, 10 );
   cr_assert( px->agg_.status_ == PC_STATUS_UNKNOWN );
   cr_assert( px->last_slot_ == 1025 );
   cr_assert( px->num_qt_ == 0 );
+  cr_assert( px->timestamp_ == 10 );
+  cr_assert( px->prev_slot_ == 1025 );
+  cr_assert( px->prev_timestamp_ == 5 );
```
jayantk, line 652: add a test case that validates that prev_* doesn't update if the agg.status is unknown prior to calling upd_aggregate

jayantk (approved):
nice!

jayantk (approved):

PR #155, opened 2022-04-13
Check that accounts are rent exempt

tompntn:
This PR checks that the data held by each account is exactly the size we expect, and that the accounts hold enough lamports to be rent exempt.


jayantk (approved):
LGTM. 

Before we deploy, let's be absolutely sure that all of the existing accounts satisfy the new stricter checks.

jayantk (approved):

PR #159, opened 2022-04-21
Make batch size configurable

jayantk:
Add a command-line option to change the batch size. I want this to run some experiments that measure slot hit rates with different batch sizes.

I have no idea how to test this. Is the right thing to do to use the mock publisher?

also please review critically. I haven't seriously written C++ since 2006.


pcapps/pythd.cpp
```diff
@@ -112,6 +113,7 @@ int main(int argc, char **argv)
       case 'w': cnt_dir = optarg; break;
       case 'l': log_file = optarg; break;
       case 'm': cmt = str_to_commitment(optarg); break;
+      case 'b': max_batch_size = ::atoi(optarg); break;
```
tompntn, line 116: You can parse this into an `unsigned long int` (which is the same size as an `unsigned int`) using `strtoul(optarg, NULL, 10)`, which avoids the need for the cast below.

pc/manager.cpp
```diff
@@ -746,7 +758,7 @@ void manager::on_response( rpc::get_slot *res )
   if (
     has_status( PC_PYTH_RPC_CONNECTED )
   ) {
-    
+
```
tompntn, line 761: Nitpick: spurious whitespace change.

pcapps/pythd.cpp
```diff
@@ -154,6 +156,12 @@ int main(int argc, char **argv)
       << mgr.get_mapping_pub_key_file() << "]" << std::endl;
     return 1;
   }
+
+  if (max_batch_size > 0) {
+    mgr.set_max_batch_size((unsigned) max_batch_size);
+  }
+  std::cout << "pythd: max batch size " << mgr.get_max_batch_size();
```
tompntn, line 163: Missing a `<< std::endl` here.

tompntn (approved):

tompntn (approved):

PR #166, opened 2022-04-28
Filter out publisher quotes with wide CIs

jayantk:
We've had several instances where publishers have submitted prices that are far from everyone else, but with really wide CIs. The aggregation logic is robust to this case, so a single publisher doing this will not cause a problem. However, this particular problem happens frequently enough that I would like to deal with it in a better way.

In previous cases when this has happened, we have asked publishers to not provide a price if their spread / CI is too wide. E.g., the serum publisher implements this logic https://github.com/pyth-network/pyth-serum/blob/main/program/src/serum-pyth/serum-pyth.c#L266 . 

However, it's a little silly that we keep asking publishers to do this when we can do it directly in the on-chain program. This change copies that serum publisher logic into the logic for updating a publisher's price.

### DIFF
--- a/program/src/oracle/oracle.c
+++ b/program/src/oracle/oracle.c
@@ -6,7 +6,7 @@
 #include "upd_aggregate.h"
 
 // Returns the minimum number of lamports required to make an account
-// with dlen bytes of data rent exempt. These values were calculated 
+// with dlen bytes of data rent exempt. These values were calculated
 // using the getMinimumBalanceForRentExemption RPC call, and are
 // guaranteed never to increase.
 static uint64_t rent_exempt_amount( uint64_t dlen )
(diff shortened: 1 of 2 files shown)


program/src/oracle/oracle.c
```diff
@@ -561,9 +561,21 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
...
   // update component price if required
   if ( cptr->cmd_ == e_cmd_upd_price || cptr->cmd_ == e_cmd_upd_price_no_fail_on_error ) {
+    uint32_t status = cptr->status_;
+
+    // Set publisher's status to unknown unless their CI is sufficiently tight.
+    int64_t threshold_conf = (cptr->price_ / PC_MAX_CI_DIVISOR);
+    if (threshold_conf < 0) {
+      // Safe as long as threshold_conf isn't the min int64, which it isn't as long as PC_MAX_CI_DIVISOR > 1.
+      threshold_conf = -threshold_conf;
+    }
```
tompntn, line 571: I believe this is equivalent to casting the result of the division to an unsigned integer.

```suggestion
    // Safe as long as threshold_conf isn't the min int64, which it isn't as long as PC_MAX_CI_DIVISOR > 1.
    uint64_t threshold_conf = (uint64_t)(cptr->price_ / PC_MAX_CI_DIVISOR);
```

program/src/oracle/test_oracle.c
```diff
@@ -303,7 +303,7 @@ Test( oracle, add_publisher ) {
 
   // Now give the price account enough lamports to be rent exempt
   acc[1].lamports = &PRICE_ACCOUNT_LAMPORTS;
-  
+
```
tompntn, line 306: Nitpic: spurious whitespace change.

tompntn (approved):

PR #167, opened 2022-04-28
Flush batches early

jayantk:
At the moment, pythd sends all of the price update transactions at roughly the same time. I think this is causing the later transactions to fail (though I'm not 100% sure).

Before we added batching, pythd staggered price updates by staggering requests for prices. When the client responded to this request, pythd immediately sent a price update transaction. This approach had the effect of staggering the update transactions (because the staggered update request schedule is directly reflected in the TXs). We lost this staggering when we added batching, because the client responses were queued and then flushed at the start of the next slot.

This change adds back staggering by eagerly sending transactions for any complete batches during the slot. It also explicitly flushes any pending updates at the end of the price request schedule; this flushing handles the case where there aren't enough prices to complete a single batch.

### DIFF
--- a/pc/manager.cpp
+++ b/pc/manager.cpp
@@ -753,7 +753,7 @@ void manager::on_response( rpc::get_slot *res )
   if (has_status( PC_PYTH_RPC_CONNECTED ) && !is_pub_) {
     // Flush any partial batches of updates, as we've reached the end of the list of products.
     for( user *uptr = olist_.first(); uptr; uptr = uptr->get_next() ) {
-      uptr->send_pending_upds();
+      uptr->send_pending_upds(uptr->num_pending_upds());
     }
   }
 
--- a/pc/manager.cpp
+++ b/pc/manager.cpp
@@ -500,11 +500,18 @@ void manager::poll( bool do_wait )
     tconn_.reconnect();
   }
 
-  // submit new quotes while connected
   if ( has_status( PC_PYTH_RPC_CONNECTED ) &&
        !hconn_.get_is_err() &&
        ( !wconn_ || !wconn_->get_is_err() ) ) {
+    // request product quotes from pythd's clients while connected
     poll_schedule();
+
+    // send any pending complete price update batches to solana
+    for( user *uptr = olist_.first(); uptr; uptr = uptr->get_next() ) {
+      if (uptr->num_pending_upds() >= get_max_batch_size()) {
+        uptr->send_pending_upds(get_max_batch_size());
+      }
+    }
   } else {
     reconnect_rpc();
   }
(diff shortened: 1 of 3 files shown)


pc/user.cpp
```diff
@@ -330,17 +330,27 @@ void user::parse_get_product( uint32_t tok, uint32_t itok )
   add_tail( itok );
 }
 
-void user::send_pending_upds()
+uint32_t user::num_pending_upds()
+{
+  return pending_vec_.size();
+}
+
+void user::send_pending_upds(uint32_t n)
```
tompntn, line 338: Regarding testing: we don't have an easy way to reliably test _when_ transactions are sent. To test _if_ partial batches and full batches are sent, you could use something like [test_update_price.py](https://github.com/pyth-network/pyth-client/blob/main/pyth/tests/test_update_price.py).

pc/manager.cpp
```diff
@@ -743,6 +750,13 @@ void manager::on_response( rpc::get_slot *res )
     clnt_.send( breq_ );
   }
 
+  if (has_status( PC_PYTH_RPC_CONNECTED ) && !is_pub_) {
+    // Flush any partial batches of updates, as we've reached the end of the list of products.
+    for( user *uptr = olist_.first(); uptr; uptr = uptr->get_next() ) {
+      uptr->send_pending_upds(uptr->num_pending_upds());
+    }
```
tompntn, line 757: It might be nice to factor this loop out into an `manager::send_pending_upds` function.

pc/user.cpp
```diff
@@ -332,15 +335,24 @@ void user::parse_get_product( uint32_t tok, uint32_t itok )
 
 void user::send_pending_upds()
 {
-  if ( pending_vec_.empty() ) {
+  uint32_t n_sent = 0;
```
tompntn, line 338: I think `to_send` might be a clearer variable name: to me, `n_sent` implies that that it's calculating the number of updates that have already been sent.

pc/user.hpp
```diff
@@ -43,7 +43,8 @@ namespace pc
     // symbol price schedule callback
     void on_response( price_sched *, uint64_t ) override;
 
-    // send all pending updates
+    // send a batch of pending price updates. This function eagerly sends any complete batches.
```
tompntn, line 46: Might be worth clarifying the semantics:

`// If there are multiple complete batches, only one will be sent.`

tompntn: I assume those are the desired semantics, as the aim of this is to stagger transactions.

tompntn (approved):

tompntn (approved):

PR #168, opened 2022-05-04
Improve accuracy of rent exemption check

tompntn:
This ensures that the rent exemption check is correct for accounts that are larger than expected, and maintains the invariant that all our accounts are one of the expected sizes.


jayantk (approved):
🙇

PR #175, opened 2022-05-24
Batch price updates across all websocket users

wijagels:
Expand batching to apply to all JRPC clients. This allows publishers to push prices from different systems and still benefit from the massive gas savings batching can provide.

~`price::send()` never returns false, so the logic to track which users participated in a batch could be removed without much impact.~

### DIFF
--- a/pc/manager.hpp
+++ b/pc/manager.hpp
@@ -264,6 +264,12 @@ namespace pc
     rpc::get_recent_block_hash breq_[1]; // block hash request
     rpc::program_subscribe     preq_[1]; // program account subscription
     rpc::get_program_accounts  areq_[1]; // alternative to program_subscribe
+
+    // price updates to be sent as part of the next batch
+    std::vector<price*> pr_upds_;
+
+    // Timestamp of the last batch
+    int64_t previous_ts_ = 0;
   };
 
   inline bool manager::get_is_tx_connect() const
--- a/pc/manager.hpp
+++ b/pc/manager.hpp
@@ -224,8 +224,10 @@ namespace pc
     void poll_schedule();
     void reset_status( int );
 
-    // sends pending updates if the previous send was more than PC_FLUSH_INTERVAL ago
-    // or there is >= get_max_batch_size() pending updates
+    // send a batch of pending price updates. This function eagerly sends any complete batches.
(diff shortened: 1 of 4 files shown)


pc/manager.hpp
```diff
@@ -264,6 +271,12 @@ namespace pc
     rpc::get_recent_block_hash breq_[1]; // block hash request
     rpc::program_subscribe     preq_[1]; // program account subscription
     rpc::get_program_accounts  areq_[1]; // alternative to program_subscribe
+
+    // price updates to be sent as part of the next batch
+    std::vector<price*> pr_upds_;
```
tompntn, line 276: Could we keep this as `pending_upds_`?

tompntn, line 276: Could we call this `pending_upds_`?

pc/manager.hpp
```diff
@@ -264,6 +271,12 @@ namespace pc
     rpc::get_recent_block_hash breq_[1]; // block hash request
     rpc::program_subscribe     preq_[1]; // program account subscription
     rpc::get_program_accounts  areq_[1]; // alternative to program_subscribe
+
+    // price updates to be sent as part of the next batch
```
tompntn, line 275:
```suggestion
    // price updates that have not been sent yet
```
```suggestion
    // price updates to be sent as part of the next batch
```

pc/manager.hpp
```diff
@@ -264,6 +271,12 @@ namespace pc
...
+    std::vector<price*> pr_upds_;
+
+    // Timestamp of the last batch
+    int64_t previous_ts_ = 0;
```
tompntn, line 279: Could we keep this as `last_upd_ts_`?

pc/manager.hpp
```diff
@@ -221,6 +224,10 @@ namespace pc
     void poll_schedule();
     void reset_status( int );
 
+    // sends pending updates if the previous send was more than PC_FLUSH_INTERVAL ago
+    // or there is >= get_max_batch_size() pending updates
```
tompntn, line 228: Could we copy the docstring from the previous `user::send_pending_ups` function?

tompntn, line 228: Could we copy the docstring from the previous user::send_pending_ups function?

tompntn (approved):

PR #178, opened 2022-06-21
Add support for a secondary network

tompntn:
This PR adds support for publishing and subscribing to updates from a secondary network.

To do this, we instantiate another instance of the `pc::manager` class, which manages the RPC connection and account subscriptions from the secondary network.

pythd websocket API users are still managed soley by the primary manager, but the responses to the websocket requests are populated by the secondary manager's data if the primary manager does not have enough data to do so.


jayantk (dismissed):

pc/manager.cpp
```diff
@@ -535,30 +586,23 @@ void manager::poll( bool do_wait )
...
-  while ( is_pub_ && kidx_ < kvec_.size() ) {
-    price_sched *kptr = kvec_[kidx_];
-    int64_t pub_ts = pub_ts_ + static_cast< int64_t >(
-      ( static_cast< uint64_t >( pub_int_ ) * kptr->get_hash() )
-        / price_sched::fraction
-    );
-    if ( curr_ts_ > pub_ts ) {
+  for ( uint32_t i = 0; i < kvec_.size(); i++ ) {
+    price_sched *kptr = kvec_[i];
+    if ( ( curr_ts_ - kptr->get_last_scheduled_time() ) > pub_int_ ) {
```
jayantk, line 604: the old logic would space out the price scheduling requests in time. I think the goal of this logic was to space out the solana transactions to increase the probability of landing them. I'm pretty sure this new logic will request all of the prices in bursts. This may cause some issues with tx hit rate, though hard to say for sure. I think you can either demonstrate that it doesn't cause a problem, or make this logic space out the txs again.

jayantk (approved):
nice!

PR #179, opened 2022-06-28
Check exponent range

jayantk:
Bound the range on price account exponents `8 >= expo >= -8` such the the EMA computation can't lose too much precision.


ptaffet (approved):
Looks good. I didn't check all the existing prices, but this does prevent any new prices with strange exponents. The value is only used when creating a price, so I don't see any potential unintended consequences of this change.

PR #182, opened 2022-07-04
Remove upd_test and init_test oracle instructions

tompntn:
These are no longer used or needed.


Reisen (approved):
Looks good, command indices maintained as well 👍

PR #183, opened 2022-07-05
Add C++ pythd websocket example

tompntn:
This PR demonstrates how to use C++ to publish price updates to the Pyth Network through the pythd websocket API described at https://docs.pyth.network/publish-data/pyth-client-websocket-api

It's a very rough, minimal example which demonstrates how to use the API.

### DIFF
--- a/pctest/test_publish_websocket.cpp
+++ b/pctest/test_publish_websocket.cpp
@@ -162,7 +162,7 @@ void pythd_websocket::get_product_list_and_subscribe( )
   });
 
   req->connect(req.get(), &jcon::JsonRpcRequest::error, [](int code, const QString& message) {
-    std::cout << "error sending get_product_list " << message.toStdString() << std::endl;
+    std::cout << "error sending get_product_list (" << code << ") " << message.toStdString() << std::endl;
   });
 }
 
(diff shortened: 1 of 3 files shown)


pctest/test_publish_websocket.cpp
```diff
@@ -0,0 +1,268 @@
...
+  - Look up the price account associated with the subscription ID.
+  - If we have any new price information for that price account, send an update_price message to submit this
+    to the network.
+*/
+
+// This class is responsible for the communication with pythd over the websocket protocol
+// described at https://docs.pyth.network/publish-data/pyth-client-websocket-api
+// 
+// It abstracts away any solana-specific details, such as public keys.
+class pythd_websocket
```
jayantk, line 37: can we expose this as non-test code so that people can use it without copy/pasting ?

jayantk (commented):
the high-level approach makes sense to me. I left you some more tactical comments.

pctest/test_publish_websocket.cpp
tompntn: Decided against this, as

- each publisher is likely to use the api slightly differently
- publishers are unlikely to use QT as the event loop

PR #185, opened 2022-07-06
add logging to upd_price and upd_aggregate

cctdaniel:
This PR adds logging to `upd_price` and `upd_aggregate` price which makes up > 99% of pyth's transactions and hence allows easier parsing for metrics/analytics.

This will help us track certain metrics such as:
- number of daily aggregate prices pushed by pyth
- pyth price feed uptime based on status
- and more

Due to solana C sdk limitations (no libc available: https://github.com/solana-labs/solana/tree/master/sdk/bpf/c#limitations), the chosen approach is to separate the logging of static string and `uint64_t` to multiple lines instead of a single line. `sol_log_64` prints 64 bit values represented in hexadecimal to stdout and hence would have to be converted to decimal while processing the data for metrics/analytics.

Comparison of compute units before and after logging for an `UpdatePrice` instruction:
- Before
```
Transaction executed in slot 1871:
  Signature: 5VJHyZAk5qTb9Va7LEJPeVAD2TZR1eKsVMwoXSSo1DxPg4ELTs2nq3U41oK7y3TMgT4eDYJmDUCX26jqh5VrVP9t
  Status: Ok
  Log Messages:
    Program 2Uy3CybcD5kJVH7vZ4fhFUKyLWJtVZvc2cAdphoNS35M invoke [1]
    Program 2Uy3CybcD5kJVH7vZ4fhFUKyLWJtVZvc2cAdphoNS35M consumed 797 of 200000 compute units
    Program 2Uy3CybcD5kJVH7vZ4fhFUKyLWJtVZvc2cAdphoNS35M success
```

- After
```
Transaction executed in slot 4060:
  Signature: F1PMozoZqSrzmEfTXCHm6dWt5uCThvoFtiP9Pw8MDhH7sFQk28eyk5iTEzy8o6kfAC4n2utmYroZJhi3baoXKyV
  Status: Ok
  Log Messages:
    Program Ci9zRpz2DniZ6myKwc5LCE5bg2hLiuCttB296M2XFgzp invoke [1]
    Program log: Instruction: UpdatePrice
    Program log: price, conf, status, pub_slot
    Program log: 0xa412, 0x3, 0x1, 0xfdb, 0x0
    Program Ci9zRpz2DniZ6myKwc5LCE5bg2hLiuCttB296M2XFgzp consumed 954 of 200000 compute units
    Program Ci9zRpz2DniZ6myKwc5LCE5bg2hLiuCttB296M2XFgzp success
```

### DIFF
--- a/scripts/solana.patch
+++ b/scripts/solana.patch
@@ -1,5 +1,5 @@
 diff --git a/sdk/bpf/c/bpf.mk b/sdk/bpf/c/bpf.mk
-index 541629ad49..8c2ec94041 100644
+index 541629ad49..31e4f46856 100644
 --- a/sdk/bpf/c/bpf.mk
 +++ b/sdk/bpf/c/bpf.mk
 @@ -14,6 +14,12 @@ TEST_PREFIX ?= test_
@@ -25,19 +25,11 @@ index 541629ad49..8c2ec94041 100644
    -Werror \
    -O2 \
    -fno-builtin \
-@@ -68,6 +77,7 @@ BPF_CXX_FLAGS := \
-   -march=bpfel+solana
- 
- BPF_LLD_FLAGS := \
-+  -z defs \
-   -z notext \
-   -shared \
-   --Bdynamic \
-@@ -245,6 +255,7 @@ define TEST_EXEC_RULE
+@@ -245,6 +254,7 @@ define TEST_EXEC_RULE
  $1: $2
- 	LD_LIBRARY_PATH=$(TESTFRAMEWORK_RPATH) \
- 	$2$(\n)
-+	$2 $(TEST_FLAGS)$(\n)
+        LD_LIBRARY_PATH=$(TESTFRAMEWORK_RPATH) \
+        $2$(\n)
++       $2 $(TEST_FLAGS)$(\n)
  endef
  
  .PHONY: $(INSTALL_SH)
--- a/scripts/solana.patch
+++ b/scripts/solana.patch
@@ -27,9 +27,9 @@ index 541629ad49..31e4f46856 100644
    -fno-builtin \
 @@ -245,6 +254,7 @@ define TEST_EXEC_RULE
  $1: $2
-        LD_LIBRARY_PATH=$(TESTFRAMEWORK_RPATH) \
-        $2$(\n)
-+       $2 $(TEST_FLAGS)$(\n)
+ 	LD_LIBRARY_PATH=$(TESTFRAMEWORK_RPATH) \
+ 	$2$(\n)
++	$2 $(TEST_FLAGS)$(\n)
  endef
  
  .PHONY: $(INSTALL_SH)
--- a/program/c/src/oracle/oracle.c
+++ b/program/c/src/oracle/oracle.c
@@ -5,6 +5,43 @@
 #include "oracle.h"
 #include "upd_aggregate.h"
 
+
+// Returns the respective log messages given the command.
+static const char* get_log_message( command_t cmd )
+{
+  switch ( cmd )
+  {
+  case e_cmd_init_mapping:
+    return "Instruction: InitMapping";
+  case e_cmd_add_mapping:
+    return "Instruction: AddMapping";
+  case e_cmd_add_product:
+    return "Instruction: AddProduct";
+  case e_cmd_upd_product:
+    return "Instruction: UpdateProduct";
+  case e_cmd_add_price:
+    return "Instruction: AddPrice";
+  case e_cmd_add_publisher:
+    return "Instruction: AddPublisher";
+  case e_cmd_del_publisher:
+    return "Instruction: DeletePublisher";
+  case e_cmd_upd_price:
+    return "Instruction: UpdatePrice";
+  case e_cmd_agg_price:
+    return "Instruction: AggregatePrice";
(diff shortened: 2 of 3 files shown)


program/src/oracle/oracle.c
```diff
@@ -510,6 +510,9 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
     fptr->conf_     = cptr->conf_;
     fptr->status_   = status;
     fptr->pub_slot_ = cptr->pub_slot_;
+    sol_log("Instruction: UpdatePrice");
+    sol_log("price, conf, status, pub_slot");
+    sol_log_64(fptr->price_ , fptr->conf_, fptr->status_, fptr->pub_slot_, 0);
```
cctdaniel, line 515: `sol_log_64()` takes 5 args and so I passed the last args as 0, what do you think about logging _ as an unused variable so that it's clearer that the last var is unused? or do you think this is fine? @jayantk @tompntn

program/src/oracle/oracle.c
```diff
@@ -510,6 +510,9 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
     fptr->conf_     = cptr->conf_;
     fptr->status_   = status;
     fptr->pub_slot_ = cptr->pub_slot_;
+    sol_log("Instruction: UpdatePrice");
```
tompntn, line 513: It might be nicer to log the `hdr->cmd_` enum value in the `dispatch` function, so that we log this for all instructions.

tompntn (commented):
Nice!

program/src/oracle/oracle.c
```diff
@@ -510,6 +510,9 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
...
+    sol_log("price, conf, status, pub_slot");
```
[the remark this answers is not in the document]
cctdaniel: I'm planning to parse the logs and look for `Instruction: UpdatePrice` for all price update txs and the reason why this format is chosen is because other major protocol like Mango and Drift also does this, happy to consider other alternatives, as for `price, conf, status, pub_slot` is to make it clear that the values below is of such order, we could of course remove it and just document it somewhere in our gitbook but I thought this will improve the UX for other protocols especially if they wish to compose on us or build metrics/analytics using our data

program/src/oracle/oracle.c
guibescos: YES!

program/src/oracle/oracle.c
```diff
@@ -490,6 +490,9 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
   // update aggregate price as necessary
   if ( sptr->slot_ > pptr->agg_.pub_slot_ ) {
     upd_aggregate( pptr, sptr->slot_, sptr->unix_timestamp_ );
+    sol_log("Instruction: UpdateAggregatePrice");
```
jayantk, line 493: the wording of this log statement is a little confusing to me because the instruction is actually UpdatePrice, and this just happened to be the first invocation of that within the slot, so aggregation was triggered.

program/src/oracle/oracle.c
jayantk: I like @tompntn 's suggestion. One way to get around the int issue is to make a table mapping the ints -> the expected log string.

scripts/solana.patch
```diff
@@ -25,15 +25,7 @@ index 541629ad49..8c2ec94041 100644
    -Werror \
    -O2 \
    -fno-builtin \
-@@ -68,6 +77,7 @@ BPF_CXX_FLAGS := \
```
jayantk, line 28: why did this change?

program/src/oracle/oracle.c
jayantk: and also maybe the log below should be agg_price, agg_conf, etc?

program/src/oracle/oracle.c
```diff
@@ -490,6 +529,11 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
   // update aggregate price as necessary
   if ( sptr->slot_ > pptr->agg_.pub_slot_ ) {
     upd_aggregate( pptr, sptr->slot_, sptr->unix_timestamp_ );
+    if (pptr->agg_.status_ == PC_STATUS_TRADING) {
```
jayantk, line 532: why if status==TRADING? I think it would be nice to know when the status has other values as well

jayantk (commented):
I would like to understand what's going on with the flags a bit better. aside from that lgtm.

program/src/oracle/oracle.c
cctdaniel: the initial thought was that the aggregate price will be useful only when its valid but we can definitely remove the check

scripts/solana.patch
guibescos: We're actually only changing `-z defs`. Thus the `BPF_LLD_FLAGS` is no longer modified, so it doesn't appear in the patch file.

jhunsaker: the point of this patch was to ensure that the program can be loaded by the runtime

I believe we already had an issue where pyth was down due to having an unresolved symbol

PR #190, opened 2022-07-18
changed c return value of update price to indicate aggregation

majabbour:
This is used so that rust code knows whether aggregation happened


docker/Dockerfile
```diff
@@ -26,7 +26,10 @@ RUN apt-get install -qq \
   qtchooser \
   qt5-qmake \
   qtbase5-dev-tools \
-  libqt5websockets5-dev
+  libqt5websockets5-dev\
```
guibescos, line 29: Did you add this?

guibescos (changes requested):

program/c/src/oracle/oracle.c
```diff
@@ -511,6 +512,9 @@ static uint64_t upd_price( SolParameters *prm, SolAccountInfo *ka )
     fptr->status_   = status;
     fptr->pub_slot_ = cptr->pub_slot_;
   }
+  if (updated_aggregate){
+    return SUCCESSFULLY_UPDATED_AGGREGATE;
+  }
   return SUCCESS;
 }
```
guibescos, line 519: You need to Update upd_price_no_fail as well

majabbour: Done

PR #191, opened 2022-07-19
changed c return value of update price to indicate aggregation (#190)

majabbour:
* changed c return value of update price to indicate aggregation

* forgot to add bindings

* removed the need for casting

* fixed update_no_fail

* removed unnecessary docker packages


majabbour:
Our trackers only want to track updates to the aggregate price. In order to do so, they need to be able to tell apart upd_price querries that affect the aggregate from those who do not. 

To help us with this, we also generated bindings for oracle.h in using bindgen in the buildscript. These bindings will be crucial for Rust to read the parts of the accounts that are managed by c

program/c/src/oracle/oracle.h
```diff
@@ -6,6 +6,13 @@
 extern "C" {
 #endif
 
+//a new custom return value to indicate to rust that aggregate was updated
+//this triggers SMA trackers to update
+//values 0-14 are defined in solana_sdk.h
+//used consts instead of define because bingen always turns
+// defines to u32 (even with ULL suffix)
+const uint64_t SUCCESSFULLY_UPDATED_AGGREGATE = 15ULL;
```
jayantk, line 14: The comment should be written in a timeless way, because you don't know when someone will read it. E.g., "new" isn't going to make sense a year in the future. I would stick with something like "A return value indicating that the aggregate price was updated" for the first sentence.

I suggest using a number >> 15 in order to ensure that the solana_sdk still has extensibility. what if they want to add an additional code?

program/rust/src/lib.rs
```diff
@@ -1,11 +1,19 @@
+//this is auto generated by build_bpf.sh
```
jayantk, line 1: clarify the comment: "this" meaning the whole file?

program/rust/src/lib.rs
```diff
@@ -1,11 +1,19 @@
...
     fn c_entrypoint(input: *mut u8) -> u64;
 }
 
 #[no_mangle]
 pub extern "C" fn entrypoint(input: *mut u8) -> u64 {
-    unsafe{
-        return c_entrypoint(input);
+    let c_ret_val = unsafe{c_entrypoint(input)};
+    if c_ret_val == c_oracle_header::SUCCESSFULLY_UPDATED_AGGREGATE{
+        return 0;
```
jayantk, line 13: i take it there's no way to reference SUCCESS here right? maybe leave a comment that this is SUCCESS?

jayantk (commented):
Please do define success as something >> 15, but lgtm aside from that.

program/rust/src/lib.rs
majabbour: The only way is by generating bindings for solana_sdk.h (which I encountered a lot of errors doing, and is probably an overkill), or copying their enums over to the file we are generating bingens for (this sounds not robust to Solana updates unless we copy them in a programatic way, or add to some to do list that people have to do when bumping solana version).

guibescos (approved):

PR #193, opened 2022-07-19
Derive borsch for bingen

majabbour:
We need to derive Borsh traits to C bindings, so I wrote a python script that takes the names of the structs we want to derive them for, and adds it to the bindings.

The script can be enhanced by matching the regex pub(spaces)struct(spaces)struct_name(spaces){. Instead of being strict. Though I do not see why bindgen would add extra spaces.


guibescos:
Great job! Let's fix the changes but this came out really clean.

guibescos (commented):
This needs to get rebased since the other branch got merged

PR #194, opened 2022-07-20
Borsch traits using bindgen

majabbour:


majabbour:
We need to derive Borsh traits to C bindings, so I changed build.rs to include a map from type names to traits that we can use to add traits to types in the C bindings. I included the output file in c_oracle_header.rs below imports needed to define the added traits.

PR #196, opened 2022-07-21
Borsch traits using bindgen

majabbour:
We need to derive Borsh traits to C bindings, so I changed build.rs to include a map from type names to traits that we can use to add traits to types in the C bindings. I included the output file in c_oracle_header.rs below imports needed to define the added traits.


program/rust/build.rs
```diff
@@ -1,3 +1,21 @@
...
+    let borsh_derives: Vec<String> = vec!["BorshSerialize".to_string(), "BorshDeserialize".to_string()];
+
+    let parser = build_utils::DeriveAdderParserCallback{types_to_traits: HashMap::from([
+        ("cmd_hdr", borsh_derives)
+        //map more struct names to traits here, be sure that the definitions of your traites
+        //are included in c_oracle_header.rs
+    ])};
+
+    //generate and write bindings
+    let bindings = bindgen::Builder::default().header("./src/bindings.h").parse_callbacks(Box::new(parser)).generate().expect("Unable to generate bindings");
```
tompntn, line 19: nitpic: would be good to run `rustfmt` on all Rust code in this repo as part of the CI/CD to get consistent formatting.

program/rust/build.rs
```diff
@@ -1,3 +1,21 @@
...
 fn main() {
     println!("cargo:rustc-link-search=../c/target");
+    
+    let borsh_derives: Vec<String> = vec!["BorshSerialize".to_string(), "BorshDeserialize".to_string()];
+
+    let parser = build_utils::DeriveAdderParserCallback{types_to_traits: HashMap::from([
+        ("cmd_hdr", borsh_derives)
+        //map more struct names to traits here, be sure that the definitions of your traites
+        //are included in c_oracle_header.rs
+    ])};
```
tompntn, line 16: It might be nice to put this in a function called `register_traits` or something similar, with a comment explaining why this is necessary.

program/rust/build_utils.rs
```diff
@@ -0,0 +1,25 @@
...
+//this is required to implement the callback trait
+impl UnwindSafe for DeriveAdderParserCallback<'_> {}
+
+impl bindgen::callbacks::ParseCallbacks for DeriveAdderParserCallback<'_>{
+    fn add_derives(&self, _name: &str) -> Vec<String>{
+        let traits = self.types_to_traits.get(_name);
+        match traits{
+            Some(trait_names)=>trait_names.to_vec(),
+            None => vec![],
+        }
```
tompntn, line 23: Could maybe use [unwrap_or_default](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default) here.

program/rust/build.rs
majabbour: oh sorry I have not been doing that for the code I have been pushing, I should probably make a different PR with just cargo fmt after merging this/ other outstanding rust PRs. Otherwise git diff might be confusing

majabbour: I added a all to fmt in  to the Builder in build.rs. Did not seem to change how bindings.rs  lookd much, but  I do not think it's a big problem since you should never have to read it (because it just redefines stuff in oracle.h).

I will start calling cargo fmt on every commit once we make that fmt PR

program/rust/build_utils.rs
majabbour: the trait `std::default::Default` is not implemented for `&Vec<String>`

I am not sure how to get around this because we can not define traits in a file different from where the class is defined AFAIK

majabbour: used `unwrap_or`

program/rust/build.rs
```diff
@@ -1,3 +1,18 @@
...
+use bindgen::Builder;
+use std::vec::Vec;
+
+
 fn main() {
     println!("cargo:rustc-link-search=../c/target");
+    
+    let borsh_derives: Vec<String> = vec!["BorshSerialize".to_string(), "BorshDeserialize".to_string()];
+
+    //make a parser and to it type, traits pairs
```
jayantk, line 11: i think people usually put a space after the // and before the text. Actually I'm kind of surprised that rustfmt doesn't fix this for you.

program/rust/src/lib.rs
```diff
@@ -1,7 +1,23 @@
-//c_oracle_header is auto generated by build_bpf.sh
-//to reflect the current status of oracle.h
 mod c_oracle_header;
 
+//Below is a high lever description of the rust/c setup.
```
jayantk, line 3: 🙇  for writing documentation explaining how this works.

jayantk (commented):
Looks good to me -- definitely better to rely on rust to add these traits. I'll let @guibescos give you the final approval though.

program/rust/build.rs
majabbour: yes, sorry noob Rust developer here. Forgot to run cargo fmt. Once this and the other PR are approved, I will do it in a seperate PR that is just fmt. (and then run it after every commit).

If I run it now the diffs will probably be confusing for reviewers.

guibescos (approved):

PR #197, opened 2022-07-21
Syncing account sizes between c and rust

majabbour:
While our C code does not need to directly interact with SMA trackers/ other rust objects. They need to know their sizes to check that the account is rent free, and to anticipate the right account size in pythd.

In this PR, I define the size as a constant  in oracle.h, and I add a unit test in Rust that checks that the size is correct (to make sure that the dockerimage does not build if the size is not correct).


program/c/src/oracle/oracle.h
```diff
@@ -13,6 +13,10 @@ extern "C" {
 // defines to u32 (even with ULL suffix)
 const uint64_t SUCCESSFULLY_UPDATED_AGGREGATE = 1000ULL;
 
+//The size of the time machine defined in rust
+//used to infer the size of price accounts
```
tompntn, line 17:
```suggestion
// The size of the "time machine" account defined in the 
// Rust portion of the codebase.
```

program/c/src/oracle/oracle.h
```diff
@@ -13,6 +13,10 @@ extern "C" {
...
+const uint64_t TIME_MACHINE_SIZE = 1864ULL;
```
tompntn, line 18:
```suggestion
const uint64_t TIME_MACHINE_ACCOUNT_SIZE = 1864ULL;
```

tompntn (approved):

tompntn (approved):

PR #198, opened 2022-07-21
Rust logging 2

guibescos:
Logging for Dune consumption and to allow tracking of unpermissioned publisher's data.


jayantk (approved):
nice. doing this in rust is way nicer than doing it in C

cctdaniel:
not sure how important this is but generally solana protocols log their instructions in PascalCase e.g.
- Drift -> OpenPosition, ClosePosition, DepositCollateral
- Mango -> PlaceSpotOrder2, PlacePerpOrder2, SettleFunds

so maybe we want to follow PascalCase as well? e.g. UpdatePrice, UpdatePriceNoFailOnError, AddPrice, AddPublisher, etc

PR #200, opened 2022-07-21
Called cargo fmt

majabbour:
I will make sure to call cargo fmt after every commit starting now!


tompntn (approved):

tompntn:
Could we add this to the CI/CD check?

PR #201, opened 2022-07-21
Make c robust to price account size

majabbour:
We plan on resizing price accounts to fit the Time Machine. This PRs ensures that this does not cause existing C code to fail.
There are two places where the size of price_t not being the same as the account size can cause failures:

1-rent exempt check while validating accounts (oracle.c) 
2-Decompressing the result of the RPC request to Solana requesting to read price contracts (pythd, request.cpp)

(We do not change the default new account size for Pyth-manager because it is helpful to have accounts from the old size in unit tests, and C pyth-manager is no longer used anyway)

### DIFF
--- a/pc/request.cpp
+++ b/pc/request.cpp
@@ -354,7 +354,7 @@ void product::update( T *res )
     return;
   }
   pc_prod_t *prod;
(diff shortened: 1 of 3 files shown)


guibescos:
Definitely want a second pair of eyes on this

PR #204, opened 2022-07-22
Intercepting price updates

majabbour:
Our design involves rersizing price accounts, as well as intercepting calls to instructions that update_the price in order to do the SMA tracking. 

In this PR:
1. We intercept the add_price calls (but we do not change their behavior yet)
2. We add a no-op update_account_version instruction that is meant to update the version number of a given account, and resize price accounts.

The purpose of this is to get feedback on the structure of the program.

It also addressed two other issues:

1. added panic and heap definitions. (These are usually added in the entrpoint! macro is solana sdk, code won't compile without them once we start using anything that code panic like arrays, etc) 
2. Updated the error message in the size of time machine assert so that it prints the correct size for better developer experience

### DIFF
--- a/program/rust/src/lib.rs
+++ b/program/rust/src/lib.rs
@@ -8,8 +8,6 @@ use crate::log::{post_log, pre_log};
 use solana_program::entrypoint::deserialize;
 use borsh::{BorshDeserialize, BorshSerialize};
 use solana_program::entrypoint::deserialize;
-use solana_program::pubkey::Pubkey;
-use solana_program::sysvar::slot_history::AccountInfo;
 
 
 //Below is a high lever description of the rust/c setup.
--- a/program/rust/src/lib.rs
+++ b/program/rust/src/lib.rs
@@ -51,7 +51,10 @@ pub extern "C" fn entrypoint(input: *mut u8) -> u64 {
         _ => {}
     }
 
-    let c_ret_val = process_instruction(program_id, &accounts, instruction_data, input);
+    let c_ret_val = match process_instruction(program_id, &accounts, instruction_data, input) {
+        Err(error) => error.into(),
+        Ok(success_status) => success_status,
+    };
 
     match post_log(c_ret_val, &accounts) {
         Err(error) => return error.into(),
--- a/program/rust/src/lib.rs
+++ b/program/rust/src/lib.rs
@@ -1,10 +1,19 @@
-mod c_oracle_header;
+pub mod c_oracle_header;
+mod rust_oracle;
 mod time_machine_types;
+<<<<<<< HEAD
 mod error;
 mod log;
 
 use crate::log::{post_log, pre_log};
 use solana_program::entrypoint::deserialize;
+=======
+use borsh::{BorshDeserialize, BorshSerialize};
+use solana_program::entrypoint::deserialize;
+use solana_program::pubkey::Pubkey;
+use solana_program::sysvar::slot_history::AccountInfo;
+
+>>>>>>> 4656a2b (add a new insctruction and interecepted update price calls)
 
 //Below is a high lever description of the rust/c setup.
 
@@ -45,7 +54,32 @@ pub extern "C" fn entrypoint(input: *mut u8) -> u64 {
         _ => {}
     }
 
-    let c_ret_val = unsafe { c_entrypoint(input) };
+    let cmd_hdr_size = ::std::mem::size_of::<c_oracle_header::cmd_hdr>();
+    if instruction_data.len() < cmd_hdr_size {
+        panic!("insufficient data, could not parse instruction");
+    }
+
+    let cmd_data = c_oracle_header::cmd_hdr::try_from_slice(&instruction_data[..cmd_hdr_size]).unwrap();
+
+    if cmd_data.ver_ != c_oracle_header::PC_VERSION {
+        //FIXME: I am not sure what's best to do here (this is copied from C)
+        // it seems to me like we should not break when version numbers change
+        //instead we should maintain the update logic accross version in the
+        //upd_account_version command
+        panic!("incorrect version numbers");
+    }
+
+    let c_ret_val = match cmd_data.cmd_ as u32 {
+        c_oracle_header::command_t_e_cmd_upd_price
+        | c_oracle_header::command_t_e_cmd_upd_price_no_fail_on_error
+        | c_oracle_header::command_t_e_cmd_agg_price => {
+            rust_oracle::update_price(program_id, accounts, instruction_data, input)
+        }
+        c_oracle_header::command_t_e_cmd_upd_account_version => {
+            rust_oracle::update_version(program_id, accounts, instruction_data)
+        }
+        _ => unsafe { return c_entrypoint(input) },
+    }
 
     match post_log(c_ret_val, &accounts) {
         Err(error) => return error.into(),
--- a/program/rust/src/lib.rs
+++ b/program/rust/src/lib.rs
@@ -1,19 +1,16 @@
 pub mod c_oracle_header;
 mod rust_oracle;
 mod time_machine_types;
-<<<<<<< HEAD
 mod error;
 mod log;
 
 use crate::log::{post_log, pre_log};
 use solana_program::entrypoint::deserialize;
-=======
 use borsh::{BorshDeserialize, BorshSerialize};
 use solana_program::entrypoint::deserialize;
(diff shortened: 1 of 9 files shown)


majabbour:
Our design involves rersizing price accounts, as well as intercepting calls to instructions that update_the price in order to do the SMA tracking. 

In this PR:
1. We intercept the add_price calls (but we do not change their behavior yet)
2. We add a no-op update_account_version instruction that is meant to update the version number of a given account, and resize price accounts.

The purpose of this is to get feedback on the structure of the program.

It also addressed two other issues:

1. added panic and heap definitions. (These are usually added in the entrpoint! macro is solana sdk, code won't compile without them once we start using anything that code panic like arrays, etc) 
2. Updated the error message in the size of time machine assert so that it prints the correct size for better developer experience

program/rust/src/lib.rs
```diff
@@ -33,11 +36,34 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
-    if c_ret_val == c_oracle_header::SUCCESSFULLY_UPDATED_AGGREGATE {
-        //0 is the SUCCESS value for solana
-        return 0;
-    } else {
-        return c_ret_val;
+    let (program_id, accounts, instruction_data) = unsafe { deserialize(input) };
+    let cmd_hdr_size = ::std::mem::size_of::<c_oracle_header::cmd_hdr>();
+    if instruction_data.len() < cmd_hdr_size {
+        panic!("insufficient data, could not parse instruction");
+    }
```
guibescos, line 43: Instead of panicking can we a more conventional Solana error handling. Checkout the logging branch so see how to.

program/rust/src/lib.rs
```diff
@@ -33,11 +36,34 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
 #[no_mangle]
 pub extern "C" fn entrypoint(input: *mut u8) -> u64 {
-    let c_ret_val = unsafe { c_entrypoint(input) };
-    if c_ret_val == c_oracle_header::SUCCESSFULLY_UPDATED_AGGREGATE {
-        //0 is the SUCCESS value for solana
-        return 0;
-    } else {
-        return c_ret_val;
+    let (program_id, accounts, instruction_data) = unsafe { deserialize(input) };
+    let cmd_hdr_size = ::std::mem::size_of::<c_oracle_header::cmd_hdr>();
```
guibescos, line 40: `use std::mem::size_of;` on top for better style.

program/rust/src/lib.rs
```diff
@@ -33,11 +36,34 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
+    if instruction_data.len() < cmd_hdr_size {
+        panic!("insufficient data, could not parse instruction");
+    }
+
+    let cmd_data =
+        c_oracle_header::cmd_hdr::try_from_slice(&instruction_data[..cmd_hdr_size]).unwrap();
```
guibescos, line 46: I think here it's probably better to just do `use c_oracle_header::*` and have access to all the structs.

program/rust/src/rust_oracle.rs
```diff
@@ -0,0 +1,33 @@
+use super::c_entrypoint;
+use super::c_oracle_header;
+use solana_program::pubkey::Pubkey;
+use solana_program::sysvar::slot_history::AccountInfo;
+
+///Calls the c oracle update_price, and updates the Time Machine if needed
+pub fn update_price(
```
guibescos, line 7: What about having this and `update_version` return a `ProgramResult` for better error handling.

program/rust/src/lib.rs
```diff
@@ -33,11 +36,34 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
+
+    if cmd_data.ver_ != c_oracle_header::PC_VERSION {
+        //FIXME: I am not sure what's best to do here (this is copied from C)
```
guibescos, line 49: We need to discuss what is the best way to handle the version field.

guibescos (commented):
Let's use rust error handling and better import style.

program/rust/src/lib.rs
majabbour: I think it's good to have a clear visual seperation between C and Rust stuff. but happy to change it if you are sure.
use * still sounds not great to be me, but maybe use se c_oracle_header::{val1, val2 ..}

program/rust/src/rust_oracle.rs
majabbour: `pub extern "C" fn entrypoint(input: *mut u8) -> u64 ` does not return program resunt, and neither does c_entry_point.

We would have to do something similar to entrypoint! to make the first work, and we would have to wrap c_entrypooint with something so that it also does. IDK how worth it is, I think we should prioritize having an MVP to test, what do you think?

program/rust/src/lib.rs
majabbour: We agreed on slack that this will be handled in a future PR

program/rust/src/lib.rs
```diff
@@ -38,27 +47,47 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
+    let cmd_data = cmd_hdr::try_from_slice(&instruction_data[..cmd_hdr_size]).unwrap();
+
+    if cmd_data.ver_ != PC_VERSION {
+        //FIXME: I am not sure what's best to do here (this is copied from C)
+        // it seems to me like we should not break when version numbers change
+        //instead we should log a message that asks users to call update_version
+        panic!("incorrect version numbers");
+    }
+
+    let c_ret_val = match cmd_data.cmd_ as u32 {
```
guibescos, line 67: Unsafe cast?

program/rust/src/lib.rs
```diff
@@ -38,27 +47,47 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
...
-    match pre_log(&accounts,instruction_data) {
+    match pre_log(&accounts, instruction_data) {
         Err(error) => return error.into(),
         _ => {}
     }
 
-    let c_ret_val = unsafe { c_entrypoint(input) };
+    let cmd_hdr_size = size_of::<cmd_hdr>();
+    let cmd_data = cmd_hdr::try_from_slice(&instruction_data[..cmd_hdr_size]).unwrap();
+
```
guibescos, line 59: Using unwrap is bad practice because it will panic and the user won't be able to know where the error happened. You can solve this by copying the pattern in log.rs for error propagation.

program/rust/src/rust_oracle.rs
```diff
@@ -0,0 +1,27 @@
+use super::c_entrypoint;
```
guibescos, line 1: This is gonna sound very nitpicky. People would understand better if this is called `processor.rs` and has a function `process(program_id, accounts, instruction_data) ` which is basically the match statement on `cmd_data.cmd_`. In `lib.rs` you would import it.

program/rust/src/rust_oracle.rs
guibescos: I would probably solve this by returning a `Result<u64, OracleError/ProgramError>`

program/rust/src/rust_oracle.rs
majabbour: Yes, that looks more similar to other contracts. Thank you!

program/rust/src/lib.rs
```diff
@@ -36,29 +42,43 @@ pub extern "C" fn c_entrypoint(input: *mut u8) -> u64 {
     0 //SUCCESS value
 }
 
+pub fn c_entrypoint_wrapper(input: *mut u8) -> OracleResult {
+    //Throwing an exception from C into Rust is undefined behavior
+    //This seems to be the best we can do
+    match unsafe { c_entrypoint(input) } {
+        0 => Ok(0), // Success
+        SUCCESSFULLY_UPDATED_AGGREGATE => Ok(SUCCESSFULLY_UPDATED_AGGREGATE),
+        2 => Err(ProgramError::InvalidArgument), //2 is ERROR_INVALID_ARGUMENT in solana_sdk.h
```
guibescos, line 51: Can you use the version from the bindings instead of the magic number 2?

guibescos (approved):

program/rust/src/lib.rs
majabbour: No, unless I make bindings of solana_sdk

PR #212, opened 2022-07-28
Log time and ema

guibescos:
Adding extra logs for the logs based vs account based comparison.

### DIFF
--- a/program/rust/src/log.rs
+++ b/program/rust/src/log.rs
@@ -2,8 +2,10 @@ use crate::c_oracle_header::*;
 use crate::error::OracleError;
 use borsh::BorshDeserialize;
 use solana_program::account_info::AccountInfo;
+use solana_program::clock::Clock;
 use solana_program::entrypoint::ProgramResult;
 use solana_program::msg;
+use solana_program::sysvar::Sysvar;
 use std::mem::size_of;
 
 pub fn pre_log(accounts: &[AccountInfo], instruction_data: &[u8]) -> ProgramResult {
@@ -14,11 +16,14 @@ pub fn pre_log(accounts: &[AccountInfo], instruction_data: &[u8]) -> ProgramResu
         .cmd_
         .try_into()
         .map_err(|_| OracleError::Generic)?;
+
+    let clock = Clock::get()?;
+
     match instruction_id {
         command_t_e_cmd_upd_price | command_t_e_cmd_agg_price => {
             let instruction: cmd_upd_price = cmd_upd_price::try_from_slice(instruction_data)?;
             msg!(
-                "UpdatePrice: publisher={:}, price_account={:}, price={:}, conf={:}, status={:}, slot={:}",
+                "UpdatePrice: publisher={:}, price_account={:}, price={:}, conf={:}, status={:}, slot={:}, solana_time={:}",
                 accounts.get(0)
                 .ok_or(OracleError::Generic)?.key,
                 accounts.get(1)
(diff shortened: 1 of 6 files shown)


program/rust/src/log.rs
```diff
@@ -81,24 +100,51 @@ pub fn pre_log(accounts : &[AccountInfo], instruction_data: &[u8]) -> ProgramRes
...
+        .map_err(|_| OracleError::Generic)?;
+
+        let expo: i32 = i32::try_from_slice(
+            &accounts
+                .get(1)
+                .ok_or(OracleError::Generic)?
+                .try_borrow_data()?[expo_start..(expo_start + size_of::<i32>())],
+        )?;
+
+        let clock = Clock::get()?;
```
tompntn, line 135: This could be inlined.

program/rust/src/log.rs
```diff
@@ -81,24 +100,51 @@ pub fn pre_log(accounts : &[AccountInfo], instruction_data: &[u8]) -> ProgramRes
...
+
+        let expo_start : usize = PRICE_T_EXPO_OFFSET.try_into()
+        .map_err(|_| OracleError::Generic)?;
+
+        let expo: i32 = i32::try_from_slice(
+            &accounts
+                .get(1)
+                .ok_or(OracleError::Generic)?
+                .try_borrow_data()?[expo_start..(expo_start + size_of::<i32>())],
+        )?;
```
tompntn, line 133: Could we factor these into seperate functions? To keep the size of `post_log` down.

program/rust/src/log.rs
```diff
@@ -81,24 +100,51 @@ pub fn pre_log(accounts : &[AccountInfo], instruction_data: &[u8]) -> ProgramRes
...
+                .try_borrow_data()?[ema_start..(ema_start + size_of::<pc_ema>())],
+        )?;
+
+        let expo_start : usize = PRICE_T_EXPO_OFFSET.try_into()
+        .map_err(|_| OracleError::Generic)?;
+
+        let expo: i32 = i32::try_from_slice(
+            &accounts
+                .get(1)
+                .ok_or(OracleError::Generic)?
```
tompntn, line 131: Instead of generic errors, could we create actual errors here? You can use the `anyhow:anyhow` macro to create "ad-hoc" errors from strings: https://docs.rs/anyhow/latest/anyhow/macro.anyhow.html

program/rust/src/log.rs
```diff
@@ -2,45 +2,64 @@ use crate::c_oracle_header::*;
...
         .map_err(|_| OracleError::Generic)?;
+
+    let clock = Clock::get()?;
+
+    let expo_start : usize = PRICE_T_EXPO_OFFSET.try_into()
+    .map_err(|_| OracleError::Generic)?;
+
+    let expo: i32 = i32::try_from_slice(
+        &accounts
+            .get(1)
```
tompntn, line 27: Could you explain the reason for using index 1 in a comment?

program/rust/src/log.rs
guibescos: Done

program/rust/src/log.rs
guibescos: Made a function for this in `deserialize.rs`

program/rust/src/log.rs
guibescos: Made actual errors in the enum

program/rust/src/log.rs
guibescos: Done

tompntn (approved):

PR #219, opened 2022-08-03
Migrate init_mapping to Rust

jayantk:
Migrate the init_mapping instruction from C to Rust. I started doing this as a warmup for deleting product/price accounts, but on further reflection it's probably worth migrating all this stuff over to Rust.

I had to do a couple things to make this work:
* Add Pod to the various account structs
* Expose various #defines as Rust values
* Reimplement some utility functions from C in rust
* Add some utilities for reading account data using Pod


program/rust/src/rust_oracle.rs
```diff
@@ -23,5 +33,94 @@ pub fn update_version(
...
     panic!("Need to merge fix to pythd in order to implement this");
-    Ok(0) //SUCCESS
+    Ok(SUCCESS)
 }
+
+
+/// initialize the first mapping account in a new linked-list of mapping accounts
+/// accounts[0] funding account           [signer writable]
+/// accounts[1] new mapping account       [signer writable]
+pub fn init_mapping(
```
jayantk, line 43: please check that this is a line-by-line translation of the C

program/rust/src/rust_oracle.rs
```diff
@@ -23,5 +33,94 @@ pub fn update_version(
...
+    mapping_account.magic_ = PC_MAGIC_V;
+    mapping_account.ver_   = hdr.ver_;
+    mapping_account.type_  = PC_ACCTYPE_MAPPING_V;
+    mapping_account.size_  = (size_of::<pc_map_table_t>() - size_of_val( &mapping_account.prod_ )) as u32;
+
+    Result::Ok(SUCCESS)
+}
+
+// FIXME: this is an extremely scary way to check errors because if you forget the ? after calling it, it doesn't do anything.
+pub fn pyth_assert(condition: bool, error_code: ProgramError) -> Result<(), ProgramError> {
```
jayantk, line 75: thoughts on how to make this less error prone? I think we need some error-checking utilities, but this approach scares me.

guibescos: I think clippy catches this.

majabbour: I think if we turn it into a macro we can be sure it returns an error.
I am also looking into how we can add a clippy check that all errors have been propagated into out CI/CD.

program/rust/src/c_oracle_header.rs
```diff
@@ -9,6 +9,43 @@ use borsh::{
...
+    Pod,
+    Zeroable,
+};
+
 //bindings.rs is generated by build.rs to include
 //things defined in bindings.h
 include!("../bindings.rs");
+
+#[cfg(target_endian = "little")]
+unsafe impl Zeroable for pc_map_table {
```
guibescos, line 22: I think Pod should be derived for safety, let me look into that.

program/rust/src/rust_oracle.rs
jayantk: clippy gives you a warning, but i don't think that's good enough. I'll look into making a macro as a follow up pr.

guibescos: Clippy will catch it with the deny warnings option, which is on.

program/rust/src/rust_oracle.rs
```diff
@@ -23,5 +50,117 @@ pub fn update_version(
...
+        .map_err(|_| ProgramError::InvalidArgument)?;
+    sol_memset(*data, 0, size_of::<pc_map_table_t>());
+
+    let mut mapping_account = load_account_as_mut::<pc_map_table_t>(&accounts.get(1).unwrap())
+        .map_err(|_| ProgramError::InvalidArgument)?;
+    mapping_account.magic_ = PC_MAGIC_V;
+    mapping_account.ver_ = hdr.ver_;
+    mapping_account.type_ = PC_ACCTYPE_MAPPING_V;
+    mapping_account.size_ =
+        (size_of::<pc_map_table_t>() - size_of_val(&mapping_account.prod_)) as u32;
```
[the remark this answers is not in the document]
jayantk: i agree that makes sense if failure is possible. that said, this cast is guaranteed to succeed

program/rust/src/rust_oracle.rs
jayantk: i fiddled with this a bit per the comments, so please check again. it's a little harder to tell that it's a line-by-line translation now

program/rust/src/bindings.h
```diff
@@ -13,3 +13,12 @@ typedef unsigned long int uint64_t;
 
 #include "../../c/src/oracle/oracle.h"
 #include "price_t_offsets.h"
+
+// Store #define constants as values so they are accessible from rust.
```
majabbour, line 17: I believe they should be accessible regardless, I am guessing you are doing this to control the type?

jayantk: hm the defines didn't work for me before, but it worked this time. thanks!

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+}
+
+/// Interpret the bytes in `data` as a value of type `T`
+fn load<T: Pod>(data: &[u8]) -> &T {
+    from_bytes(&data[0..size_of::<T>()])
+}
+
+/// Interpret the bytes in `data` as a mutable value of type `T`
+#[allow(unused)]
+fn load_mut<T: Pod>(data: &mut [u8]) -> &mut T {
```
majabbour, line 132: Is there another version of size_of that we might need to use here in the case of structs that have padding?

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+    from_bytes_mut(&mut data[0..size_of::<T>()])
```
majabbour, line 133: Use` try_from_bytes_mut` instead of `from_bytes_mut`, the latter panics, which makes debugging harder

program/rust/src/rust_oracle.rs
jayantk: not sure what exactly you're asking here. I think the slice below handles any padding in the allocated account

majabbour: nvm size_of is the right thing to use (see link below)
https://github.com/Lokathor/bytemuck/blob/main/src/internal.rs#L171

majabbour: (for refrence if we are using borsh, we would need to use their size function as it is different for types whose size is not a multiple of 8)

program/rust/src/rust_oracle.rs
guibescos: In the load functions I think it can panic if data is too short, which can happen for instructions.

guibescos: Maybe use `data.get(0..size_of::<T>())` instead of `data[0..size_of::<T>()]`

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+        .try_borrow_mut_data()
+        .map_err(|_| ProgramError::InvalidArgument)?;
+    let length = data.len();
+    sol_memset(data.borrow_mut(), 0, length);
+    Ok(())
+}
+
+/// Interpret the bytes in `data` as a value of type `T`
+fn load<T: Pod>(data: &[u8]) -> &T {
+    from_bytes(&data[0..size_of::<T>()])
```
majabbour, line 127: `try_from_bytes` instead for error propagation

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+fn load_mut<T: Pod>(data: &mut [u8]) -> &mut T {
+    from_bytes_mut(&mut data[0..size_of::<T>()])
+}
+
+/// Get the data stored in `account` as a value of type `T`
+fn load_account_as<'a, T: Pod>(account: &'a AccountInfo) -> Result<Ref<'a, T>, ProgramError> {
+    let data = account.try_borrow_data()?;
+
+    Ok(Ref::map(data, |data| {
+        bytemuck::from_bytes(&data[0..size_of::<T>()])
```
majabbour, line 141: `try_from_bytes` instead for error propagation

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+
+/// Mutably borrow the data in `account` as a value of type `T`.
+/// Any mutations to the returned value will be reflected in the account data.
+fn load_account_as_mut<'a, T: Pod>(
+    account: &'a AccountInfo,
+) -> Result<RefMut<'a, T>, ProgramError> {
+    let data = account.try_borrow_mut_data()?;
+
+    Ok(RefMut::map(data, |data| {
+        bytemuck::from_bytes_mut(&mut data[0..size_of::<T>()])
```
majabbour, line 153: `try_from_bytes_mut` instead for error propagation

program/rust/src/rust_oracle.rs
majabbour: If you do not use try, it would panic if data is too short or too long. it needs to be an exact match

guibescos: Actually I think the way you solve this is by checking the size of instruction data in the dispatcher

program/rust/src/rust_oracle.rs
```diff
@@ -23,4 +51,105 @@ pub fn update_version(
...
+        }
+        _ => Err(ProgramError::InvalidArgument),
+    }?;
+
+    // Initialize by setting to zero again (just in case) and populating the account header
+    clear_account(fresh_mapping_account)?;
+
+    let hdr = load::<cmd_hdr_t>(instruction_data);
+    {
+        let mut mapping_account = load_account_as_mut::<pc_map_table_t>(fresh_mapping_account)?;
```
guibescos, line 82: Rename `mapping_account` to` `mapping_data` ?

program/rust/src/rust_oracle.rs
jayantk: for this one, i couldn't figure out how to swap the Ref<Result<...>> to Result<Ref<..>>

majabbour: I took a quick look, the only thing I could find is this:
https://stackoverflow.com/questions/72730465/how-do-i-propagate-an-error-outside-of-a-closure
Though not sure if it is worth the hastle, I think Guillermo or Tom might be more helpful. Specially Guillermo since he had a strong opinion on unwrap

I will check again tomorrow though

guibescos (approved):

PR #220, opened 2022-08-04
Add price rust

guibescos:
Add price instruction in Rust.


program/rust/src/rust_oracle.rs
```diff
@@ -87,6 +95,68 @@ pub fn init_mapping(
...
+/// accounts[1] product account to add the price account to       [signer writable]
+/// accounts[2] newly created price account                       [signer writable]
+pub fn add_price(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> OracleResult {
+    if instruction_data.len() < size_of::<cmd_add_price_t>() {
+        return Err(ProgramError::InvalidArgument);
+    }
```
jayantk, line 109: i think we should put this check into `load`

program/rust/src/rust_oracle.rs
```diff
@@ -87,6 +95,68 @@ pub fn init_mapping(
...
+    }?;
+
+    let mut product_data = load_account_as_mut::<pc_prod_t>(product_account)?;
+    {
+        let price_data = load_account_as::<pc_price_t>(price_account)?;
+
+        if product_data.magic_ != PC_MAGIC
+            || product_data.ver_ != cmd_args.ver_
+            || product_data.type_ != PC_ACCTYPE_PRODUCT
+            || price_data.magic_ != 0
```
jayantk, line 137: One thing i've noticed from doing this translation exercise is that the check for whether an account is uninitialized differs in all of the instructions. `init_mapping` checks both the magic and ver fields, but `add_account` only checks the magic. I think it should be fine for us to standardize on one of these and encode that in `validate_fresh_account`.

jayantk (commented):
lgtm, will approve once you merge/rebase

program/rust/src/rust_oracle.rs
```diff
@@ -102,21 +112,64 @@ pub fn add_mapping(
...
         ProgramError::InvalidArgument,
     )?;
 
     initialize_mapping_account(next_mapping, hdr.ver_)?;
-    unsafe {
-        cur_mapping
-            .next_
-            .k1_
-            .copy_from_slice(&next_mapping.key.to_bytes());
+    pubkey_assign(cur_mapping.next_, &next_mapping.key.to_bytes());
```
jayantk, line 123: 🙇

jayantk (dismissed):
nice

jayantk (approved):

PR #221, opened 2022-08-04
Implement add_mapping instruction

jayantk:
As it says. I pulled out some helpers to read accounts & check their content, and also to initialize new accounts of a type. I think we can follow this pattern of functions for the other types of accounts as well.

### DIFF
--- a/program/rust/src/rust_oracle.rs
+++ b/program/rust/src/rust_oracle.rs
@@ -26,6 +26,7 @@ use crate::c_oracle_header::{
     pc_map_table_t,
     PC_ACCTYPE_MAPPING,
     PC_MAGIC,
+    PC_MAP_TABLE_SIZE,
 };
 use crate::error::OracleResult;
 
@@ -77,7 +78,7 @@ pub fn init_mapping(
     }?;
 
     // Initialize by setting to zero again (just in case) and populating the account header
-    let hdr = load::<cmd_hdr_t>(instruction_data);
+    let hdr = load::<cmd_hdr_t>(instruction_data)?;
     initialize_mapping_account(fresh_mapping_account, hdr.ver_)?;
 
     Ok(SUCCESS)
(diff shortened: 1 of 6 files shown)


program/rust/src/rust_oracle.rs
```diff
@@ -75,14 +78,44 @@ pub fn init_mapping(
...
     let hdr = load::<cmd_hdr_t>(instruction_data)?;
-    let mut mapping_data = load_account_as_mut::<pc_map_table_t>(fresh_mapping_account)?;
-    mapping_data.magic_ = PC_MAGIC;
-    mapping_data.ver_ = hdr.ver_;
-    mapping_data.type_ = PC_ACCTYPE_MAPPING;
-    mapping_data.size_ = (size_of::<pc_map_table_t>() - size_of_val(&mapping_data.prod_)) as u32;
+    let mut cur_mapping = load_mapping_account_mut(cur_mapping, hdr.ver_)?;
+    pyth_assert(
+        cur_mapping.num_ == PC_MAP_TABLE_SIZE
+            && unsafe { cur_mapping.next_.k8_.iter().all(|x| *x == 0) },
```
jayantk, line 108: we need the `unsafe` blocks here because the account keys are a union type.

guibescos: What about this syntax : 
```
cur_mapping.next_.k1_ = next_mapping.key.to_bytes());
```

guibescos (approved):

program/rust/src/rust_oracle.rs
```diff
@@ -150,3 +183,38 @@ fn load_account_as_mut<'a, T: Pod>(
...
+        ProgramError::InvalidArgument,
+    )?;
+
+    Ok(mapping_account_ref)
+}
+
+/// Initialize account as a new mapping account. This function will zero out any existing data in
+/// the account.
+fn initialize_mapping_account(account: &AccountInfo, version: u32) -> Result<(), ProgramError> {
+    clear_account(account)?;
```
majabbour, line 210: I wonder if it's a good idea to only keep instructions in oracle.rs, and move helpers into other modules. (It currently looks very clean, am just imagining the file will grow a lot as we migrate more instructions)

guibescos: This is bound to happen at some time. Arguably each instruction could have its own file also.

jayantk: we can move stuff around later. i agree this file is going to get too big, but it's not yet clear to me what the right split into multiple files is.

PR #222, opened 2022-08-05
Test init mapping

guibescos:
Rewrote the test from test_oracle.c . Now we can delete that test and init_mapping from oracle.c.


jayantk (approved):
awesome!

PR #224, opened 2022-08-05
Removed Borsh

majabbour:
Since we are planning to use `bytemuck` for all `c_repr` structs,  I removed borsh and all the things we did because of borsh limitation (like passing member offsets).

I kept the structure for automotacally adding derives to bindgen structs in case we need that for something else.


jayantk (approved):
nice this is so much better

program/rust/src/log.rs
```diff
@@ -15,8 +14,7 @@ use solana_program::sysvar::Sysvar;
 pub fn pre_log(accounts: &[AccountInfo], instruction_data: &[u8]) -> ProgramResult {
     msg!("Pyth oracle contract");
 
-    let instruction_header: cmd_hdr =
-        deserialize_single_field_from_buffer::<cmd_hdr>(instruction_data, None)?;
+    let instruction_header: cmd_hdr = *load::<cmd_hdr>(instruction_data)?;
```
guibescos, line 17: Can nitpick you change these to 
`let instruction_header: &cmd_hdr = load::<cmd_hdr>(instruction_data)?;` (no dereference).
No other lines should need to change.

guibescos (changes requested):

guibescos (approved):

PR #228, opened 2022-08-08
Migrate upd_product instruction

jayantk:
This instruction was slightly tricky because of how the list of string key-value pairs is handled in the C. I made some minor changes to the implementation logic to simplify it, but they shouldn't change the semantics. Please do check that this looks right to you.

I also added my own test because we didn't have one before.


program/rust/src/tests/test_upd_product.rs
```diff
@@ -0,0 +1,163 @@
...
+    assert!(account_has_key_values(&product_account, &kvs).unwrap());
+
+    // bad size on the string
+    instruction_data[0] = 7;
+    assert!(upd_product(
+        &program_id,
+        &[funding_account.clone(), product_account.clone()],
+        &instruction_data[..size]
+    )
+    .is_err());
```
guibescos, line 91: I think it's good to compare to the actual error i.e. 
`assert_eq(upd_product(
        &program_id,
        &[funding_account.clone(), product_account.clone()],
        &instruction_data[..size]
    ), Err(ProgramError::Invalid)`

program/rust/src/tests/test_upd_product.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    let kvs = ["foo", "barz"];
+    let size = populate_instruction(&mut instruction_data, &kvs);
+    assert!(upd_product(
+        &program_id,
+        &[funding_account.clone(), product_account.clone()],
+        &instruction_data[..size]
+    )
+    .is_ok());
+    assert!(account_has_key_values(&product_account, &kvs).unwrap());
```
guibescos, line 72: `.is_ok()` instead of unwrap?

program/rust/src/tests/test_upd_product.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    // bad size on the string
+    instruction_data[0] = 7;
```
guibescos, line 85: Not sure what this test is accomplishing. You're changing the instruction discriminator.

guibescos (changes requested):

program/rust/src/tests/test_upd_product.rs
guibescos: Ok this is fine, you're actually checking the bool. I would still use `unwrap_or`

guibescos (approved):

PR #229, opened 2022-08-08
Rust/add publisher

guibescos:
Adding publishers in rust.


program/rust/src/rust_oracle.rs
```diff
@@ -170,6 +175,57 @@ pub fn add_price(
...
+/// accounts[0] funding account                                   [signer writable]
+/// accounts[1] price account to add the publisher to             [signer writable]
+pub fn add_publisher(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> OracleResult {
+    let cmd_args = load::<cmd_add_publisher_t>(instruction_data)?;
+
+    if instruction_data.len() != size_of::<cmd_add_publisher_t>() || pubkey_is_zero(&cmd_args.pub_)
```
jayantk, line 188: use pyth_assert for these checks

program/rust/src/rust_oracle.rs
```diff
@@ -170,6 +175,57 @@ pub fn add_price(
...
+    {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    let [_funding_account, price_account] = match accounts {
```
jayantk, line 193: _ prefix indicates the variable isn't used later in the function, but _funding_account is

program/rust/src/rust_oracle.rs
```diff
@@ -170,6 +175,57 @@ pub fn add_price(
...
+    check_valid_funding_account(_funding_account)?;
+    check_valid_signable_account(program_id, price_account, size_of::<pc_price_t>())?;
+
+    let mut price_data = load_price_account_mut(price_account, cmd_args.ver_)?;
+
+    if price_data.num_ >= PC_COMP_SIZE {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    for i in 0..(price_data.num_ as usize) {
```
jayantk, line 207: (i'm going to look at whether we can override some of the field definitions in the struct so they're usize instead of u32 so we don't have to do these casts all the time)

jayantk (commented):
looks good to me, just a couple nitpicks.

jayantk (approved):

PR #230, opened 2022-08-08
New interface with time machine

majabbour:
This PR implements the resize instruction, as well as the new update price logic, without implementing the TwapTracker (which is currently left as a byte array with no_op operations).

I would appreciate feedback on:
1- the interface between rust_oracle.rs and the time_machine
2- The way I compined all the into in price_t into one struct
3- The way I dealt with integer conversion
4- The resize instruction test. (Assuming it is combined with unit_tests for add_first_price)


program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +39,91 @@ pub struct TimeMachineWrapper {
...
+    //an instance of the c price_t type
+    pub price_data:            pc_price_t,
+    //space for more publishers
+    pub extra_publisher_space: [u8; EXTRA_PUBLISHER_SPACE as usize],
+    //TimeMachine
+    pub time_machine:          TimeMachineWrapper,
+}
+impl PriceAccountWrapper {
+    #[cfg(feature = "resize-account")]
+    pub fn add_first_price(&mut self) -> Result<(), OracleError> {
```
jayantk, line 79: I think this function and the one below should verify that the current status == TRADING

program/rust/src/time_machine_types.rs
```diff
@@ -1,4 +1,36 @@
...
+use crate::error::OracleError;
+use bytemuck::{
+    Pod,
+    Zeroable,
+};
+use solana_program::msg;
+
+pub trait Tracker {
+    /// add the first price to a zero initialized tracker
+    fn add_first_price(
```
jayantk, line 14: maybe call this `initialize`?

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +39,91 @@ pub struct TimeMachineWrapper {
...
+    pub time_machine:          TimeMachineWrapper,
+}
+impl PriceAccountWrapper {
+    #[cfg(feature = "resize-account")]
+    pub fn add_first_price(&mut self) -> Result<(), OracleError> {
+        self.time_machine.add_first_price(
+            self.price_data
+                .timestamp_
+                .try_into()
+                .map_err(|_| OracleError::IntegerCastingError)?,
```
jayantk, line 84: use `try_convert` ? it's a little more ergonomic

program/rust/src/time_machine_types.rs
```diff
@@ -1,4 +1,36 @@
...
+
+    ///add a new price to a tracker
+    fn add_price(
+        &mut self,
+        current_time: u64,
+        prev_time: u64,
+        current_price: i64,
+        prev_price: i64,
+        current_conf: u64,
+        prev_conf: u64,
```
jayantk, line 29: super nitpicky, but i think it would be better to sort the arguments like:

prev_time, 
prev_price
prev_conf,
current_time,
current_price,
current_conf,

this is likely easier for someone to remember because (1) previous comes before current, and (2) all the arguments for each are in a single block

program/rust/src/rust_oracle.rs
```diff
@@ -44,27 +56,138 @@ use crate::utils::pyth_assert;
...
-    //For now, we did not change the behavior of this. this is just to show the proposed structure
-    // of the program
-    c_entrypoint_wrapper(input)
+    let price_account_info = &accounts[1];
+    let account_len = price_account_info.try_data_len()?;
+    if account_len < PRICE_ACCOUNT_SIZE {
+        if account_len != PRICE_T_SIZE {
+            return Err(ProgramError::InvalidArgument);
+        }
+        msg!("Please resize the account to allow for SMA tracking!");
```
jayantk, line 76: can we standardize on SMA or TWAP instead of using both terms

jayantk: (I prefer SMA since we're using EMA already)

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +39,91 @@ pub struct TimeMachineWrapper {
...
+        _prev_price: i64,
+        _current_conf: u64,
+        _prev_conf: u64,
+    ) -> Result<(), OracleError> {
+        msg!("implement me");
+        Ok(())
+    }
+}
+
+#[derive(Copy, Clone)]
```
jayantk, line 66: you probably don't want `Copy` right? we likely want every piece of code to point to the instance of this struct that's in the solana account? if you derive copy, you create the room for bugs because the struct got copied under the hood.

pyth/tests/test_publish.py
```diff
@@ -19,6 +29,52 @@ def get_price_acct():
         output = output.decode('ascii')
         output = json.loads(output)
         return output['price_accounts'][0]
+    
+    def resize_account(price_acc_address):
+        """
+        given a string with the pubkey of a price accountm it calls the resize instruction of the Oracle on it
```
jayantk, line 35: typo price account

pyth/tests/test_publish.py
```diff
@@ -19,6 +29,52 @@ def get_price_acct():
...
+        key_data = json.load(key_file)
+        key_file.close()
+
+        sender = Keypair.from_secret_key(key_data)
+        txn.sign(sender)
+        solana_client.send_transaction(txn, sender)
+
+    def get_account_size(acc_address):
+        """
+        given a string with the pubkey of an account, resize it
```
jayantk, line 70: comment seems wrong

docker/Dockerfile
```diff
@@ -66,8 +70,11 @@ RUN cp -a /usr/bin/sdk solana
 RUN ./pyth-client/scripts/patch-solana.sh
 
 # Build and test the oracle program.
-RUN cd pyth-client && ./scripts/build-bpf.sh .
+RUN cd pyth-client && ./scripts/build-bpf.sh . "test"
```
jayantk, line 73: it's confusing to run the tests on a different version of the code than the one that gets released. What happens if the tests pass on the version with the feature flag enabled, but fail on the other version? 

I think you're doing this to test your code, but it would be better to find a different way to run the tests.

scripts/build-bpf.sh
```diff
@@ -28,7 +28,7 @@ set -x
 cd "${C_DIR}"
 export V="${V:-1}"
 make clean 
-make  "${@:2}" 
+make
```
jayantk, line 31: are you sure that you can drop the extra arguments to `make` here?

jayantk (commented):
I left you a bunch of comments inline. The rust code is generally fine and the comments are minor. I am a little worried about the change to the build scripts / dockerfile though (see the inline comments). I think you're doing this in order to run the python tests, but it would be better to do this in a way that doesn't touch the production build pipeline.

program/rust/src/rust_oracle.rs
```diff
@@ -44,27 +56,138 @@ use crate::utils::pyth_assert;
...
+    }
 }
 /// has version number/ account type dependant logic to make sure the given account is compatible
 /// with the current version
 /// updates the version number for all accounts, and resizes price accounts
+/// accounts[0] funding account           [signer writable]
+/// accounts[1] upgradded acount       [signer writable]
+/// accounts [2] system program
+#[cfg(feature = "resize-account")]
+pub fn update_version(
```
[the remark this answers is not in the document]
majabbour: The idea originally is that this is the generic upgrade account to the current version instruction. Happy to change the name though

program/rust/src/rust_oracle.rs
```diff
@@ -44,27 +56,138 @@ use crate::utils::pyth_assert;
...
+) -> OracleResult {
+    pyth_assert(
+        valid_funding_account(&funding_account_info),
+        OracleError::InvalidFundingAccount.into(),
+    )?;
+
+    pyth_assert(
+        valid_writable_account(program_id, &price_account_info, size_of::<pc_price_t>()),
+        OracleError::InvalidWritableAccount.into(),
+    )?;
```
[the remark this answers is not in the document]
majabbour: We could! 
The impression that I got talking from thomaz is that he prefers to not to have to deal with the private keys when we do not need to.

docker/Dockerfile
guibescos: Yeah the reason for this is that the resizing feature is disabled in production

program/rust/src/rust_oracle.rs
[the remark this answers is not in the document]
guibescos: Agree with making permissioned

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +39,91 @@ pub struct TimeMachineWrapper {
     place_holder: [u8; 1864],
 }
 
+impl Tracker for TimeMachineWrapper {
```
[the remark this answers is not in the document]
majabbour: There will be!
The individual SMA and Tick trackers in the wrapper will all be Trackers

scripts/build-bpf.sh
majabbour: There was no argument being provided. I think this was just left there from an older version of the build-script.
I will double  check that with an echo on the older version and trying to build the docker image. (though am also sure this properly builds the C)

docker/Dockerfile
majabbour: fair point, I recall that you and I and Guillermo talked about this, and we said we wanted to have a different "feature" for resizing accounts.
Should I run the tests twice? that would take ~40 minutes. or should maybe run them on the production version, then redeploy the "test" version and just test the resize and publish instruction?

program/rust/src/time_machine_types.rs
majabbour: sure!

program/rust/src/rust_oracle.rs
majabbour: I do not believe I am using TWAP in my part, do you main rename things in the C- code?
I do recall having a discussion with Jeff, and him having a strong prefrence for "SMA"

program/rust/src/time_machine_types.rs
majabbour: Right, I do not need to check the previous status but it is always the last with trading status, but for the current one, there's no such gurantee, thanks

program/rust/src/rust_oracle.rs
```diff
@@ -44,27 +56,138 @@ use crate::utils::pyth_assert;
...
+
+    let account_len = price_account_info.try_data_len()?;
+    //const PRICE_T_SIZE: usize = size_of::<pc_price_t>();
+    match account_len {
+        PRICE_T_SIZE => {
+            //compute the number of lamports needed in the price account to update it
```
tompntn, line 110: This comment is slightly misleading - it should maybe be "ensure account is still rent exempt after resizing".

program/rust/src/rust_oracle.rs
```diff
@@ -44,27 +56,138 @@ use crate::utils::pyth_assert;
...
+                invoke(
+                    &transfer_instruction,
+                    &[
+                        funding_account_info.clone(),
+                        price_account_info.clone(),
+                        system_program.clone(),
+                    ],
+                )?;
+            }
+            //resize
```
tompntn, line 131: Could we extract this transfer into a seperate function? To make the size of this method smaller.

pyth/tests/test_publish.py
```diff
@@ -49,8 +105,47 @@ def get_price_acct():
...
     assert after['publisher_accounts'][0]['price'] == 150
     assert after['publisher_accounts'][0]['conf'] == 7
     assert after['publisher_accounts'][0]['status'] == 'trading'
+
+    resize_account(pyth_init_price['LTC'])
+    time.sleep(20)
+    #defined in oracle.h
+    new_account_size = 6176
+    assert get_account_size(pyth_init_price['LTC']) == new_account_size
+    time.sleep(20)
```
tompntn, line 118: Is this extra sleep necessary?

docker/Dockerfile
tompntn: Could we run them twice, but configure the GitHub Actions workflow to do it in parallel?

program/rust/src/time_machine_types.rs
majabbour: Copy is required by bytmuck

program/rust/src/rust_oracle.rs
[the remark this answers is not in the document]
majabbour: I think I will name it upgrade_account, so that you can pass accounts through it after any oracle upgrade

program/rust/src/time_machine_types.rs
majabbour: I made so that add_price verifies that things have been initialized in addition to that

program/rust/src/time_machine_types.rs
majabbour: Actually, I changed the interface a little bit, lmk what you think of the new name!

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
+/// accounts[1] upgradded acount       [signer writable]
+/// accounts [2] system program
+pub fn upgrade_account(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
     _instruction_data: &[u8],
 ) -> OracleResult {
-    panic!("Need to merge fix to pythd in order to implement this");
-    // Ok(SUCCESS)
+    let funding_account_info = &accounts[0];
```
guibescos, line 164: Can you use similar syntax to the other rust instructions?

program/rust/src/time_machine_types.rs
```diff
@@ -1,4 +1,35 @@
...
+pub trait Tracker {
+    /// Makes the tracker invalidate the entry it's currentrly at
+    /// Must be used when we start tracking so that we do not give a "valid"
+    /// entry that is not completely covered by our prices
+    fn invalidate_current_entry(&mut self, last_valid_price_time: u64) -> Result<(), OracleError>;
+
+    ///add a new price to a tracker
+    fn add_price(
+        &mut self,
+        prev_time: u64,
```
guibescos, line 23: i64

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +38,95 @@ pub struct TimeMachineWrapper {
...
+    //an instance of the c price_t type
+    pub price_data:            pc_price_t,
+    //space for more publishers
+    pub extra_publisher_space: [u8; EXTRA_PUBLISHER_SPACE as usize],
+    //TimeMachine
+    pub time_machine:          TimeMachineWrapper,
+}
+impl PriceAccountWrapper {
+    pub fn invalidate_current_time_machine_entries(&mut self) -> Result<(), OracleError> {
+        let last_valid_price_time = if self.price_data.agg_.status_ == PC_STATUS_TRADING {
```
guibescos, line 78: This are all implementation details for another PR.

program/rust/src/rust_oracle.rs
jayantk: ok i thought i saw TWAP somewhere else in the pr. Definitely no need to go around renaming extra stuff as part of this.

(ah, your PR description says TwapTracker)

jayantk: also suggest not logging a message here, as it will look like an error in all the solana program logs. possible bad publicity if we release this.

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
+        price_account.add_price_to_time_machine()?;
+        msg!("updated tracker!");
+    }
+    Ok(c_ret_value)
+}
+fn send_lamports<'a>(
+    from: &AccountInfo<'a>,
+    to: &AccountInfo<'a>,
+    system_program: &AccountInfo<'a>,
+    ammount: u64,
```
jayantk, line 99: typo

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
     _instruction_data: &[u8],
     input: *mut u8,
 ) -> OracleResult {
-    //For now, we did not change the behavior of this. this is just to show the proposed structure
-    // of the program
-    c_entrypoint_wrapper(input)
+    let price_account_info = &accounts[1];
+    let account_len = price_account_info.try_data_len()?;
+    if account_len < PRICE_ACCOUNT_SIZE {
+        if account_len != PRICE_T_SIZE {
```
guibescos, line 81: Match statement

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
+            let mut price_account = load_account_as_mut::<PriceAccountWrapper>(price_account_info)?;
+            price_account.invalidate_current_time_machine_entries()?;
+            Ok(SUCCESS)
+        }
+        PRICE_ACCOUNT_SIZE => Ok(SUCCESS),
+        _ => Err(ProgramError::InvalidArgument),
+    }
+}
+
+fn get_account_type(account: &AccountInfo) -> Result<u32, ProgramError> {
```
jayantk, line 148: nitpick: this doesn't seem like it's worth a helper function -- it's the same number of lines of code to write the one line body inline.

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
+    let transfer_instruction = transfer(from.key, to.key, ammount);
+    invoke(
+        &transfer_instruction,
+        &[from.clone(), to.clone(), system_program.clone()],
+    )?;
+    Ok(())
+}
+
+///resize price account and initialize the TimeMachineStructure
+pub fn upgrade_price_account<'a>(
```
jayantk, line 110: I think this doesn't need to be public

program/rust/src/rust_oracle.rs
```diff
@@ -32,47 +40,154 @@ use crate::c_oracle_header::{
...
+/// We should call this instruction on any account that needs to be
+/// upgraded after an Oracle update
 /// updates the version number for all accounts, and resizes price accounts
-pub fn update_version(
-    _program_id: &Pubkey,
-    _accounts: &[AccountInfo],
+/// accounts[0] funding account           [signer writable]
+/// accounts[1] upgradded acount       [signer writable]
+/// accounts [2] system program
+pub fn upgrade_account(
```
jayantk, line 159: fwiw i'm not sure it makes sense to have a generic `upgrade_account` instruction. We don't do account upgrades often, and this isn't saving us much code even if we add another account upgrade in the future.

Given that you already wrote it, this is fine, but figured i'd share my thoughts.

program/rust/src/tests/test_init_mapping.rs
```diff
@@ -75,7 +76,7 @@ fn test_init_mapping() {
             &[funding_account.clone(), mapping_account.clone()],
             instruction_data
         ),
-        Err(ProgramError::InvalidArgument)
+        Err(OracleError::InvalidFundingAccount.into())
```
jayantk, line 79: note on pr structure: you shouldn't have changed these error codes at the same time as the rest of the PR. if you sent a PR with just this error change, we would insta-approve it.

guibescos (changes requested):

program/rust/src/rust_oracle.rs
guibescos: Yep, it's going to become upgrade_price_account

program/rust/src/rust_oracle.rs
```diff
@@ -38,41 +45,119 @@ use crate::c_oracle_header::{
...
+    accounts: &[AccountInfo],
     _instruction_data: &[u8],
     input: *mut u8,
 ) -> OracleResult {
-    //For now, we did not change the behavior of this. this is just to show the proposed structure
-    // of the program
-    c_entrypoint_wrapper(input)
+    let [_funding_account_info, price_account_info, _sysvar_clock] = match accounts {
+        [x, y, z] => Ok([x, y, z]),
+        _ => Err(ProgramError::InvalidArgument),
```
guibescos, line 75: You need to fix this, it's actually more like :
```
    let [funding_account, price_account, clock_account] = match accounts {
        [x, y,z ] => Ok([x, y, z]),
        [x, y,_ , z ] => Ok([x, y, z]),
        _ => Err(ProgramError::InvalidArgument),
    }?;
```

program/rust/src/rust_oracle.rs
```diff
@@ -38,41 +45,119 @@ use crate::c_oracle_header::{
...
+    }?;
```
guibescos, line 76: I think you need to check that's a price account here : 
I recommended using `load_checked<pc_price_t>`

program/rust/src/rust_oracle.rs
```diff
@@ -38,41 +45,119 @@ use crate::c_oracle_header::{
...
+
+    check_valid_funding_account(funding_account_info)?;
+    check_valid_signable_account(program_id, price_account_info, size_of::<pc_price_t>())?;
+    pyth_assert(
+        check_id(system_program.key),
+        OracleError::InvalidSystemAccount.into(),
+    )?;
```
guibescos, line 126: I think you need to check that's a price account here : 
I recommended using `load_checked<pc_price_t>`

pyth/tests/test_publish.py
guibescos: typo

pyth/tests/test_publish.py
```diff
@@ -19,6 +29,59 @@ def get_price_acct():
...
+            keys=[
+                AccountMeta(pubkey=funding_key, is_signer=True, is_writable=True),
+                AccountMeta(pubkey=price_key, is_signer=True, is_writable=True),
+                AccountMeta(pubkey=system_key, is_signer=False, is_writable=False),
+            ],
+            program_id = PublicKey(solana_program_deploy),
+        )
+        txn = Transaction().add(resize_instruction)
+        txn.fee_payer = funding_key
+        sender = get_key_pair(solana_keygen[1])
```
guibescos, line 69: Nickpick call this `funding_key_pair`.

majabbour: Not really, the type is PublicKey.

guibescos: It's a signer to it has to be a keypair

program/rust/src/rust_oracle.rs
majabbour: There's a load_checked at the end, with a comment that explains why it is sufficient.
This is cleaner than having to accesss a smaller price account then a bigger one

pyth/tests/test_publish.py
majabbour: No, the signing happens somehwere else :)

majabbour: `solana_client.send_transaction(txn, sender, price_key_pair)`
This is where signing happens, it just needs the public key for the feepayer

majabbour: I can changer sender to fee_payer_keypair, will do in a sec

guibescos: Sorry, my comment was clear : I'm proposing renaming `sender` to `funding_key_pair`

program/rust/src/rust_oracle.rs
jayantk: Since this is a thin wrapper around `c_entrypoint_wrapper`, you want to avoid writing code that could accidentally change the behavior. The best way to have approached this problem, imo, would have been something like:

```
let c_ret_value = c_entrypoint_wrapper(input)?;

let price_account = accounts.get(2)?; 
let account_len = ...
match {
  ... => price_account.add_price_to_time_machine();
} 
```

If you structure your code that way, all of your additions happen *after* `c_entrypoint_wrapper`, which makes it hard for future edits to accidentally forget to call that function.

program/rust/src/rust_oracle.rs
```diff
@@ -38,41 +45,119 @@ use crate::c_oracle_header::{
...
+) -> Result<(), ProgramError> {
+    let transfer_instruction = transfer(from.key, to.key, amount);
+    invoke(
+        &transfer_instruction,
+        &[from.clone(), to.clone(), system_program.clone()],
+    )?;
+    Ok(())
+}
+
+/// resizes a price accpunt so that it fits the Time Machine
```
jayantk, line 107: typo: account

program/rust/src/rust_oracle.rs
jayantk: fwiw i think it's better to add a call to `load_checked` here in addition to your existing check. I agree your code as written is technically correct, but it's more prone to human error: future edits to this code could easily eliminate the necessary check.

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +22,52 @@ pub struct TimeMachineWrapper {
...
+unsafe impl Zeroable for PriceAccountWrapper {
+}
+
+#[cfg(target_endian = "little")]
+unsafe impl Pod for PriceAccountWrapper {
+}
+
+impl PythAccount for PriceAccountWrapper {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRICE;
+    const INITIAL_SIZE: u32 = size_of::<PriceAccountWrapper>() as u32;
```
guibescos, line 58: Make this consistent with PythAccount for `pc_price_t`

pyth/tests/test_publish.py
jayantk: still a typo here accountm

jayantk (commented):
I pretty much had the same comments as @guibescos .

program/rust/src/rust_oracle.rs
```diff
@@ -38,41 +45,125 @@ use crate::c_oracle_header::{
...
+    }?;
+
+    check_valid_funding_account(funding_account_info)?;
+    check_valid_signable_account(program_id, price_account_info, size_of::<pc_price_t>())?;
+    pyth_assert(
+        check_id(system_program.key),
+        OracleError::InvalidSystemAccount.into(),
+    )?;
+    //panick if not a price account
+    //need to makre sure it goes out of scope immediatly to avoid mutable borrow errors
```
guibescos, line 129: Typos

program/rust/src/time_machine_types.rs
```diff
@@ -7,10 +22,52 @@ pub struct TimeMachineWrapper {
...
+unsafe impl Zeroable for PriceAccountWrapper {
+}
+
+#[cfg(target_endian = "little")]
+unsafe impl Pod for PriceAccountWrapper {
+}
+
+impl PythAccount for PriceAccountWrapper {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRICE;
+    const INITIAL_SIZE: u32 = size_of::<pc_price_t>() as u32;
```
guibescos, line 58: PC_PRICE_T_COMP_OFFSET

guibescos (approved):
Good to go

PR #233, opened 2022-08-09
Traits2

guibescos:
The goal of this PR is to remove duplicate code in testing when instantiating Pyth accounts and the funder account.
The AccountSetup struct can provide fresh accounts of the right size and the right balance for Pyth to work on.


program/rust/src/tests/test_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+const UPPER_BOUND_OF_ALL_ACCOUNT_SIZES: usize = 20536;
+
+/// The goal of this struct is to easily instantiate zeroed solana accounts
+/// for the Pyth program to use in tests.
+pub struct AccountSetup {
+    key:     Pubkey,
+    owner:   Pubkey,
+    balance: u64,
+    size:    usize,
+    data:    [u8; UPPER_BOUND_OF_ALL_ACCOUNT_SIZES],
```
jayantk, line 18: maybe leave a comment as to why this array needs to be allocated here (I think this is to get the ownership right?)

program/rust/src/c_oracle_header.rs
```diff
@@ -18,6 +18,9 @@ include!("../bindings.rs");
 pub trait PythAccount: Pod {
     const ACCOUNT_TYPE: u32;
     const INITIAL_SIZE: u32;
+    fn minimum_size() -> usize {
```
jayantk, line 21: document the difference between INITIAL_SIZE and minimum_size. this is going to get confusing

jayantk (commented):
I like it. I asked for a couple comments, will approve once those are added

jayantk (approved):

PR #235, opened 2022-08-09
Set min pub instruction

jayantk:
I had 20 mins so knocked out this easy instruction. Added a new test again because we didn't have one in the C.

### DIFF
--- a/program/rust/src/rust_oracle.rs
+++ b/program/rust/src/rust_oracle.rs
@@ -19,7 +19,26 @@ use solana_program::program_memory::{
 use solana_program::pubkey::Pubkey;
 use solana_program::rent::Rent;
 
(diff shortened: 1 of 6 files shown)


program/rust/src/rust_oracle.rs
```diff
@@ -322,6 +323,32 @@ pub fn upd_product(
...
 
+pub fn set_min_pub(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> OracleResult {
+    let cmd = load::<cmd_set_min_pub_t>(instruction_data)?;
+
+    pyth_assert(
+        instruction_data.len() == size_of::<cmd_set_min_pub_t>(),
```
jayantk, line 334: i'm pretty sure this is unnecessary (should be covered by load), but i'm being defensive in literally copying the C.

program/rust/src/tests/test_set_min_pub.rs
```diff
@@ -0,0 +1,89 @@
...
+        true,
+        &mut price_balance,
+        &mut price_raw_data,
+        &program_id,
+        false,
+        Epoch::default(),
+    );
+
+    initialize_checked::<pc_price_t>(&price_account, PC_VERSION).unwrap();
+
```
guibescos, line 59: Looks good, can you just add a test that checks what `min_pub_` is at initialization?

guibescos (approved):

guibescos (approved):

PR #236, opened 2022-08-09
Check length

guibescos:
More graceful error handling in load. Now it shouldn't panic if length is not right.

### DIFF
--- a/program/rust/src/deserialize.rs
+++ b/program/rust/src/deserialize.rs
@@ -16,13 +16,21 @@ use solana_program::program_error::ProgramError;
 
 /// Interpret the bytes in `data` as a value of type `T`
 pub fn load<T: Pod>(data: &[u8]) -> Result<&T, ProgramError> {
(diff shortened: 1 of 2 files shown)


program/rust/src/deserialize.rs
```diff
@@ -16,13 +16,21 @@ use solana_program::program_error::ProgramError;
 
 /// Interpret the bytes in `data` as a value of type `T`
 pub fn load<T: Pod>(data: &[u8]) -> Result<&T, ProgramError> {
-    try_from_bytes(&data[0..size_of::<T>()]).map_err(|_| ProgramError::InvalidArgument)
```
jayantk, line 19: nitpick: it would be nice to document the exact error conditions here. I think this errors if there aren't enough bytes to read a T in `data`, but is tolerant of `data` containing more bytes than needed.

jayantk (approved):
great. you can add this comment in a later PR so we don't have to go through another approval / ci if you want.

PR #237, opened 2022-08-09
Rust/del publisher

guibescos:
Deleting publishers, I also wrote some generative tests to test adding-deleting but I'll leave that for a future PR.


jayantk (approved):
nice

PR #242, opened 2022-08-10
Implement init_price instruction

jayantk:


program/rust/src/c_oracle_header.rs
```diff
@@ -127,13 +126,22 @@ unsafe impl Zeroable for pc_ema {
 unsafe impl Pod for pc_ema {
 }
 
+#[cfg(target_endian = "little")]
```
jayantk, line 129: this was missing before for some reason

program/rust/src/rust_oracle.rs
```diff
@@ -169,6 +171,69 @@ pub fn add_price(
...
+        bytes_of_mut(&mut price_data.twac_),
+        0,
+        size_of::<pc_ema_t>(),
+    );
+    sol_memset(
+        bytes_of_mut(&mut price_data.agg_),
+        0,
+        size_of::<pc_price_info_t>(),
+    );
+    for i in 0..(price_data.comp_.len() as usize) {
```
guibescos, line 221: Ok, so the C code uses `price_data.num_` as the upper range. I think you're using the size of the array. In that case what about using `PC_COMP_SIZE`

program/rust/src/rust_oracle.rs
```diff
@@ -169,6 +171,69 @@ pub fn add_price(
     Ok(SUCCESS)
 }
 
+pub fn init_price(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> OracleResult {
+    let cmd_args = load::<cmd_init_price_t>(instruction_data)?;
+
```
guibescos, line 180: The check that `instruction_data.len() == size_of::<cmd_init_price_t>(cmd_init_price)` went away because `load` will work as long as `instruction_data.len() >= size_of::<cmd_init_price_t>(cmd_init_price)`

program/rust/src/tests/test_init_price.rs
```diff
@@ -0,0 +1,140 @@
...
+    assert_eq!(
+        init_price(
+            &program_id,
+            &[funding_account.clone(), price_account.clone()],
+            instruction_data
+        ),
+        Err(ProgramError::InvalidArgument)
+    );
+
+    price_account.is_signer = false;
```
guibescos, line 124: I think you meant to make this true

program/rust/src/rust_oracle.rs
guibescos: Just flagging the differences with C, I think this is acceptable.

program/rust/src/rust_oracle.rs
jayantk: right. i decided to go with len() because it seemed less brittle than `num_`. I think it's more explicit to use `len()` than PC_COMP_SIZE (as in, it's clear to the reader of the code what this does, without them having to look up the definition of PC_COMP_SIZE)

program/c/src/oracle/oracle.c
```diff
@@ -38,17 +38,6 @@ static bool valid_funding_account( SolAccountInfo *ka )
...
-static bool valid_signable_account( SolParameters *prm,
-                                    SolAccountInfo *ka,
-                                    uint64_t min_dlen )
-{
-  return ka->is_signer &&
-         ka->is_writable &&
-         SolPubkey_same( ka->owner, prm->program_id ) &&
-         ka->data_len >= min_dlen &&
-         is_rent_exempt( *ka->lamports, ka->data_len );
-}
```
jayantk, line 50: no longer used, and the c build fails if you have unused functions

guibescos (approved):

PR #243, opened 2022-08-10
Specifying account check errors

majabbour:
It's a little hard to debug when all account checks return InvalidArgument, so I specified the error dpending on which check failed


jayantk (approved):
thanks, this is a good quality of life improvement

PR #248, opened 2022-08-12
Upd price in rust

guibescos:
`upd_price` in rust. 
Also moved 3 unit tests from C to Rust.

The boundary between C and Rust becomes the `upd_aggregate` function


program/rust/src/error.rs
```diff
@@ -28,7 +28,7 @@ pub enum OracleError {
     InvalidSystemAccount    = 606,
     #[error("InvalidWritableAccount")]
     InvalidWritableAccount  = 607,
-    #[error("InvalidWritableAccount")]
+    #[error("InvalidFreshAccount")]
```
guibescos, line 31: This message was a typo

program/c/makefile
```diff
@@ -7,6 +7,6 @@ cpyth:
 	bash -c "ar rcs target/libcpyth.a target/**/*.o"
 cpythtest:
 #   Compile C code to system architecture for use by rust's cargo test
-	$(CC) -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
+	gcc -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
 #   Bundle C code compiled to system architecture for use by rust's cargo test
```
guibescos, line 11: Had to change this back... maybe $CC in the context of this makefile has the bpf flags?

jayantk: hmmm ok weird. i'm not really sure how this whole makefile process works, tbh.

program/rust/src/lib.rs
```diff
@@ -29,18 +26,18 @@ use crate::log::{
...
 
-//As we migrate from C to Rust, our Rust code needs to be able to interract with C
-//build-bpf.sh is set up to compile the C code into a bpf archive file, that build.rs
-//is set up to link the rust targets to. This enables to interact with the c_entrypoint
-//as well as similarly declare other C functions in Rust and call them
+//As we migrate from C to Rust, our Rust code needs to be able to interact with C
+//build-bpf.sh is set up to compile the C code into a two archive files
+//contained in `./program/c/target/`
+// - `libcpyth.a` contains the bpf version for production code
+// - `libcpythtest.a` contains the systems architecture version for tests
```
jayantk, line 40: fwiw would be nice if the names of these files corresponded to the architectures they were built for, like `libcpyth-bpf` and `libcpyth-native`

program/rust/src/utils.rs
```diff
@@ -123,3 +126,28 @@ pub fn read_pc_str_t(source: &[u8]) -> Result<&[u8], ProgramError> {
...
+    account: &AccountInfo,
+    minimum_size: usize,
+) -> Result<(), ProgramError> {
+    pyth_assert(
+        valid_writable_account(program_id, account, minimum_size),
+        OracleError::InvalidWritableAccount.into(),
+    )
+}
+
+pub fn is_component_update(cmd_args: &cmd_upd_price_t) -> Result<bool, ProgramError> {
```
jayantk, line 148: why "component"? 

also, the argument to this is confusing because if you have a `cmd_upd_price_t`, you would expect its cmd_ field to be the corresponding int.

program/rust/src/rust_oracle.rs
```diff
@@ -227,6 +220,118 @@ pub fn add_mapping(
...
+    }?;
+
+    check_valid_funding_account(funding_account)?;
+    check_valid_writable_account(program_id, price_account, size_of::<pc_price_t>())?;
+    // Check clock
+    let clock = Clock::from_account_info(clock_account)?;
+
+    let mut publisher_index: usize = 0;
+    let latest_aggregate_price: pc_price_info_t;
+    let latest_publisher_price: pc_price_info_t;
```
jayantk, line 247: I don't think this needs to be declared outside the block below

program/rust/src/rust_oracle.rs
```diff
@@ -227,6 +220,118 @@ pub fn add_mapping(
...
+            publisher_price.status_ = status;
+            publisher_price.pub_slot_ = cmd_args.pub_slot_;
+        }
+    }
+
+    let account_len = price_account.try_data_len()?;
+    if aggregate_updated && account_len == PRICE_ACCOUNT_SIZE {
+        let mut price_account = load_account_as_mut::<PriceAccountWrapper>(price_account)?;
+        price_account.add_price_to_time_machine()?;
+    }
```
jayantk, line 318: nitpick: I think it would be better to put this block directly under the block that computes the new aggregate. (it's just easier to follow the code that way)

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -0,0 +1,288 @@
...
+        assert_eq!(price_data.num_qt_, 0);
+        assert_eq!(price_data.timestamp_, 12);
+        assert_eq!(price_data.prev_slot_, 1025);
+        assert_eq!(price_data.prev_price_, 245);
+        assert_eq!(price_data.prev_conf_, 85);
+        assert_eq!(price_data.prev_timestamp_, 5);
+    }
+}
+
+// Create an upd_price instruction that sets the product metadata to strings
```
jayantk, line 267: comment is wrong

also why Option here for the price/conf/pub_slot?

program/rust/src/tests/test_upd_price.rs
```diff
@@ -0,0 +1,292 @@
...
+        assert_eq!(price_data.comp_[0].latest_.pub_slot_, 1);
+        assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
+        assert_eq!(price_data.valid_slot_, 0);
+        assert_eq!(price_data.agg_.pub_slot_, 1);
+        assert_eq!(price_data.agg_.price_, 0);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+    }
+
+    // add some prices for current slot - get rejected
+    populate_instruction(&mut instruction_data, Some(43), None, None);
```
jayantk, line 80: fwiw even here i think this code would be clearer if it required providing every argument. It's less state that the reader has to keep in their head.

program/rust/src/tests/test_upd_price.rs
```diff
@@ -0,0 +1,292 @@
...
+        assert_eq!(price_data.agg_.price_, 81);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
+    }
+
+    populate_instruction(&mut instruction_data, Some(50), Some(6), Some(5));
+    update_clock_slot(&mut clock_account, 6);
+
+    // Publishing a wide CI results in a status of unknown.
+
+    // check that someone doesn't accidentally break the test.
```
jayantk, line 219: I think you dropped an assert for this comment -- the point is to validate that the publisher's status is TRADING before running the instruction that sets their status to UNKNOWN. If you don't have this assertion, it's possible that a change to something before this test makes the test meaningless.

program/rust/src/tests/test_upd_price.rs
```diff
@@ -0,0 +1,292 @@
...
+        assert_eq!(price_data.comp_[0].latest_.pub_slot_, 6);
+        assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.valid_slot_, 6);
+        assert_eq!(price_data.agg_.pub_slot_, 7);
+        assert_eq!(price_data.agg_.price_, 81);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+    }
+}
+
+// Create an upd_price instruction that sets the product metadata to strings
```
jayantk, line 271: comment is wrong

program/rust/src/tests/test_upd_price_no_fail_on_error.rs
```diff
@@ -0,0 +1,154 @@
...
+        assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.valid_slot_, 0);
+        assert_eq!(price_data.agg_.pub_slot_, 1);
+        assert_eq!(price_data.agg_.price_, 0);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+    }
+
+    // Invalid updates, such as publishing an update for the current slot,
+    // should still fail silently and have no effect.
+    populate_instruction(&mut instruction_data, Some(55), Some(22), None);
```
jayantk, line 106: (this is another place where i think it would be smart to include an assert to validate that someone hasn't accidentally broken the test via a code change. I.e., check that calling `update_price` returns an error but `update_price_no_fail_on_error` doesn't.)

jayantk (commented):
awesome. Left you some minor comments

Also, are you going to delete the now unused C implementations?

program/rust/src/utils.rs
guibescos: Component as opposed to aggregate.

guibescos: This one is tricky : the dispatcher routes `upd_price`, `upd_price_no_fail_on_error` and `agg_price` to this function . The first two update the individual publisher's price, unlike `agg_price` which only tries to update the aggregate.

program/rust/src/rust_oracle.rs
guibescos: `publisher_index` and `latest_aggregate_price ` are you used afterwards. `latest_publisher_price ` can be declared inside.

program/rust/src/tests/test_upd_aggregate.rs
guibescos: Ok

program/rust/src/tests/test_upd_price.rs
guibescos: Readded the assert

program/rust/src/tests/test_upd_price_no_fail_on_error.rs
guibescos: Yes!

program/rust/src/utils.rs
jayantk: ah gotcha. this is fine

jayantk (approved):
awesome!

PR #249, opened 2022-08-12
Export upd_aggregate

guibescos:
For one of the latest steps of the migration we're going to restrict the C code to the aggregation logic.
This requires exporting the function `upd_aggregate` from C to Rust. We're doing similar logic to exporting the c_entrypoint from `oracle.c`.

A subtle difficulty is that `cargo test` runs in the system architecture, so we need to compile `upd_aggregate` to this target as well. That's the goal of `for_cargo_test/cpyth_test.c`. This will allow using `upd_aggregate` in the tests.

Of course there's danger that our rust tests pass but our production code fails because they're compiling to different targets but this problem already existed with `test_oracle.c` which is also getting compiled to the system's target. Integration tests (currently pytest) should pick this up.

### DIFF
--- a/program/c/makefile
+++ b/program/c/makefile
@@ -7,6 +7,6 @@ cpyth:
 	bash -c "ar rcs target/libcpyth.a target/**/*.o"
 cpythtest:
 #   Compile C code to system architecture for use by rust's cargo test
-	gcc -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
+	$(CC) -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
 #   Bundle C code compiled to system architecture for use by rust's cargo test
 	ar rcs target/libcpythtest.a ./target/cpyth_test.o
--- a/program/c/makefile
+++ b/program/c/makefile
@@ -3,3 +3,6 @@ SOLANA := ../../../solana
 include $(SOLANA)/sdk/bpf/c/bpf.mk
 cpyth:
 	bash -c "ar rcs target/libcpyth.a target/**/*.o"
+cpythtest:
+	gcc -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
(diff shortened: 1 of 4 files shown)


program/c/makefile
```diff
@@ -3,3 +3,6 @@ SOLANA := ../../../solana
 include $(SOLANA)/sdk/bpf/c/bpf.mk
 cpyth:
 	bash -c "ar rcs target/libcpyth.a target/**/*.o"
+cpythtest:
+	gcc -c ./src/oracle/for_cargo_test/cpyth_test.c -o ./target/cpyth_test.o
```
guibescos, line 7: I feel bad about using this "naked" `gcc` for some reason.

program/c/src/oracle/for_cargo_test/cpyth_test.c
```diff
@@ -0,0 +1,23 @@
+/// The goal of this file is to provide the upd_aggregate function to local rust tests
+
+/// The heap address is different on solana runtime and locally, so here we need to allocate some heap space
+/// In compilation to bpf, PC_HEAP_START is provided by <solana.h>
+char heap_start[8192];
+#define PC_HEAP_START (heap_start)
+#define static_assert _Static_assert
```
guibescos, line 7: I copied this from `test_oracle.c` where they also have to deal with `bpf` vs local execution

scripts/build-bpf.sh
```diff
@@ -30,13 +30,14 @@ export V="${V:-1}"
...
 make cpyth 
+make cpythtest
 rm ./target/*-keypair.json
 
 
 #build Rust and link it with C
 cd "${PYTH_DIR}"
 cargo clean
-cargo test
+cargo test-bpf
```
guibescos, line 40: I'm not sure about the difference between both but solana encourages `cargo test-bpf`

program/c/makefile
jayantk: I think you can use $CC

jayantk (approved):

PR #250, opened 2022-08-15
Implementing time machine

majabbour:
In this PR I implement the SMA and Tick tracker classes, I also provide the interface of how they will be called, and the planned testing strategy. next steps are:
1. Make sure people are ok with this interface. In particular, people seemed to dislike that these classes take current and prev prices. I believe that is ok, because they are hidden behind the add_price_to_time_machine of the price_account, and that the only alternative way to do this would be to either add a `first_price` bool to the time_machine, or duplivate members already stored in the price_account (prev_price, prev_timestamp), which sounds unnescissary if the plan is to migrate to rust and have everything be a method of the price_account
2. Discuss how many trackers we want to have and their paramteres
3. Unit test Tick and SMA trackers.
4. Make a Macro that generated the implementation of TimeMachineWrapper (since it is basically a lot of similar members, and always calling the same method on each)


program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+/// Represents an SMA Tracker that has NUM_ENTRIES entries
+/// each tracking time weighted sums for GRANUALITY seconds periods.
+///The prices are assumed to be provided under some fixed point representation, and the computation
+/// gurantees accuracy up to the last decimal digit in the fixed point representation.
+/// Assumes THRESHOLD < GRANUALITY
+pub struct SmaTracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
+    valid_entry_counter:        u64, // is incremented everytime we add a valid entry
```
guibescos, line 82: What is this being used for?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+    max_update_time:            i64, // maximum time between two updates in the current entry.
```
guibescos, line 83: Maybe rename to make clear that this is a time difference and not a timestamp?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+        Ok(())
+    }
+}
+
+impl<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64>
+    Tracker<GRANUALITY, NUM_ENTRIES, THRESHOLD> for SmaTracker<GRANUALITY, NUM_ENTRIES, THRESHOLD>
+{
+    fn initialize(&mut self) -> Result<(), OracleError> {
+        //ensure that the first entry is invalid
+        self.max_update_time = THRESHOLD;
```
guibescos, line 150: Shouldn't the first entry always get invalidated because `max_update_time = current_time - 0`?

majabbour: No, that is not the case when you resize the account!

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+    /// 2. any entry that would have been passed through during this time interval, would be
+    ///     passed through if we skip one entry at a time the number that this outputs
+    fn get_num_skipped_entries_capped(
+        prev_time: i64,
+        current_time: i64,
+    ) -> Result<usize, OracleError> {
+        let time_since_begining_of_current_entry =
+            (prev_time % GRANUALITY) + current_time - prev_time;
+        let mut num_skiped_entries: usize =
+            try_convert(time_since_begining_of_current_entry / GRANUALITY)?;
```
guibescos, line 54: Maybe easier to understand as `current_time / GRANULARITY - prev_time / GRANULARITY`

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+        //if num_skipped is greater than NUM_ENTRIES, cap it at the smallest
+        //number greater than NUM_ENTRIES that has the same residue class
+        if num_skiped_entries >= NUM_ENTRIES {
```
guibescos, line 57: If `num_skiped_entries >= NUM_ENTRIES` you're just invalidating everything so I'd be OK to just set num_skiped_entries = NUM_ENTRIES

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+/// Assumes THRESHOLD < GRANUALITY
+pub struct SmaTracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
+    valid_entry_counter:        u64, // is incremented everytime we add a valid entry
+    max_update_time:            i64, // maximum time between two updates in the current entry.
+    running_price:              [SignedTrackerRunningSum; NUM_ENTRIES], /* price running time
+                                      * weighted sums */
+    running_confidence:         [UnsignedTrackerRunningSum; NUM_ENTRIES], /* confidence running
+                                                                           * time
+                                                                           * weighted sums */
+    current_entry:              usize, //the current entry
```
guibescos, line 89: I'm not sure you need to store this, because it's just `current_time / GRANULARTIY`

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+                    if Self::get_time_to_entry_end(prev_time) >= THRESHOLD {
+                        0
+                    } else {
+                        1
+                    };
+                self.current_entry += 1;
+                self.prices[self.current_entry] = current_price;
+                self.confidences[self.current_entry] = current_conf;
+            }
+            _ => {
```
guibescos, line 315: This case is more subtle right. Because before it went down for many entries, maybe that price was close enough to the boundary.

program/rust/src/time_machine_types.rs
majabbour: At the moment it's not exactly that. But yes, we can change things so that it is  ((current_time / GRANULARTIY) % NUM_ENTRIES). Which sounds like a good idea

program/rust/src/time_machine_types.rs
```diff
@@ -35,13 +386,22 @@ pub struct PriceAccountWrapper {
...
-        Ok(())
+        self.time_machine.initialize()
     }
 
     pub fn add_price_to_time_machine(&mut self) -> Result<(), OracleError> {
-        msg!("implement me");
-        Ok(())
+        //skip if there are not two valid prices in the price account
+        if self.price_data.prev_timestamp_ == 0 {
+            return Ok(());
```
guibescos, line 395: Is this case really problematic, IMO it's just gonna be invalid anyway so we're fine.

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+    current_entry:  usize,             //the current entry
+    entry_validity: [u8; NUM_ENTRIES], //entry validity
+}
+
+
+impl<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64>
+    TickTracker<GRANUALITY, NUM_ENTRIES, THRESHOLD>
+{
+    ///invalidates current_entry, and increments self.current_entry num_entries times
+    fn invaldate_following_entries(&mut self, num_entries: usize) {
```
guibescos, line 271: Typo

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+        self.running_price[self.current_entry] +=
+            try_convert::<i64, SignedTrackerRunningSum>(update_time * avg_price)?;
+        self.running_confidence[self.current_entry] += try_convert::<u64, UnsignedTrackerRunningSum>(
+            try_convert::<i64, u64>(update_time)? * avg_conf,
+        )?;
+        //since price is always greater than confidence, it is sufficient to look at the
+        // price we also look at the average because it is the value that
+        // gets multiplied with the number of seconds
+        self.max_observed_entry_abs_val =
+            cmp::max(try_convert(avg_price)?, self.max_observed_entry_abs_val);
```
guibescos, line 140: How does this help with overflow?

program/rust/src/time_machine_types.rs
majabbour: That would break the wormhole thing, and the predicatiblity of `current_entry`. Though it would work with the next comment of ignoring `current_entry`

program/rust/src/tests/test_time_machine.txt
```diff
@@ -0,0 +1,30 @@
...
+test add_price for Tick Tracker:
+
+A (same as above)
+
+B time_from_current_entry_to_next_entry:
+less than threshohold, greater than threshold
+
+C check correct validity
+
+Taking the product of all these combinations, we get 8 tests from A and B, plus a check at the end of each test
```
guibescos, line 30: Yep just write some unit tests, sir.

guibescos (commented):
Great work overall. I think the implementation is correct and the style is good. Just dropped some comments about how to make it more understandable and debuggable. Let me know if you don't agree.

program/rust/src/time_machine_types.rs
majabbour: I think this is easier to read, but happy to replace it with the following comment:
```
//this would work on the first add_price as well, as the first entry is always invalid, so even though we would probabpy end up overflowing the first number we should be good
```

program/rust/src/time_machine_types.rs
jayantk: I don't get how this works -- it seems like valid_entry_counter continuously increments and never resets. How do i use that to determine that the intervening entries are valid?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+        //multiply by precision multiplier so that our entry aggregated can be rounded to have
+        //the same accuracy as user input
+        let inflated_prev_price = prev_price * SMA_TRACKER_PRECISION_MULTIPLIER;
+        let inflated_current_price = current_price * SMA_TRACKER_PRECISION_MULTIPLIER;
+        let inflated_prev_conf =
+            prev_conf * try_convert::<i64, u64>(SMA_TRACKER_PRECISION_MULTIPLIER)?;
+        let inflated_current_conf =
+            current_conf * try_convert::<i64, u64>(SMA_TRACKER_PRECISION_MULTIPLIER)?;
+
+        let update_time = current_time - prev_time;
```
jayantk, line 154: can you name this `update_time_delta` or something that clearly indicates that this is an interval of time, not a point?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+    valid_entry_counter:        u64, // is incremented everytime we add a valid entry
+    max_update_time_gap:        i64, // maximum time between two updates in the current entry.
+    running_price:              [SignedTrackerRunningSum; NUM_ENTRIES], /* price running time
+                                      * weighted sums */
+    running_confidence:         [UnsignedTrackerRunningSum; NUM_ENTRIES], /* confidence running
+                                                                           * time
+                                                                           * weighted sums */
+    max_observed_entry_abs_val: u64, /* used to ensure that there are no overflows when
+                                      * looking at running sum diffs */
+    entry_validity:             [u8; NUM_ENTRIES], //entry validity
```
jayantk, line 71: the `u8` here is a status code? if so, maybe make a type alias?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+pub struct SmaTracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
+    valid_entry_counter:        u64, // is incremented everytime we add a valid entry
+    max_update_time_gap:        i64, // maximum time between two updates in the current entry.
+    running_price:              [SignedTrackerRunningSum; NUM_ENTRIES], /* price running time
+                                      * weighted sums */
+    running_confidence:         [UnsignedTrackerRunningSum; NUM_ENTRIES], /* confidence running
+                                                                           * time
+                                                                           * weighted sums */
+    max_observed_entry_abs_val: u64, /* used to ensure that there are no overflows when
+                                      * looking at running sum diffs */
```
jayantk, line 70: let's say we're in a situation where this abs_val gets set high enough to break the SMA. If that happens, then we need to manually reset the state of the whole tracker to fix it?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+    entry_validity:             [u8; NUM_ENTRIES], //entry validity
+}
+
+impl<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64>
+    SmaTracker<GRANUALITY, NUM_ENTRIES, THRESHOLD>
+{
+    ///invalidates current_entry, and increments self.current_entry num_entries times
+    fn invaldate_following_entries(&mut self, num_entries: usize, current_entry: usize) {
```
jayantk, line 78: typo: invalidate

program/rust/src/time_machine_types.rs
jayantk: I think the if statement on 311 catches this case, right?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
 #[repr(C)]
 /// this wraps multiple SMA and tick trackers, and includes all the state
 /// used by the time machine
 pub struct TimeMachineWrapper {
-    //Place holder with the size of the fields I am planning to add
-    place_holder: [u8; 1864],
+    sma_tracker:
+        SmaTracker<THIRTY_MINUTES, { TEN_HOURS / (THIRTY_MINUTES as usize) }, TWENTY_SECONDS>,
+    tick_tracker:
+        TickTracker<THIRTY_MINUTES, { TEN_HOURS / (THIRTY_MINUTES as usize) }, TWENTY_SECONDS>,
```
jayantk, line 341: maybe we should do 24 hours instead of 10? that feels a little less arbitrary (because it's a whole day). we can then also roll up to longer durations, like we store 24 hour SMAs for the past 7 days.

Also, i think all of these quantities need to be measured in milliseconds (which i believe is the resolution of the timestamps)

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+        prev_price: i64,
+        prev_conf: u64,
+        current_time: i64,
+        current_price: i64,
+        current_conf: u64,
+    ) -> Result<(), OracleError> {
+        let prev_entry = Self::time_to_entry(prev_time)?;
+        let current_entry = Self::time_to_entry(current_time)?;
+        let num_skipped_entries = current_entry - prev_entry;
+        match num_skipped_entries {
```
jayantk, line 288: i find it somewhat confusing that we don't interpolate the prices between the two times in this case, but we do in the SMA case above. Is there a good reason for that? If not, I would suggest picking one or the other in both places.

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+/// Represents an SMA Tracker that has NUM_ENTRIES entries
+/// each tracking time weighted sums for GRANUALITY seconds periods.
+///The prices are assumed to be provided under some fixed point representation, and the computation
+/// gurantees accuracy up to the last decimal digit in the fixed point representation.
+/// Assumes THRESHOLD < GRANUALITY
+pub struct SmaTracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
+    valid_entry_counter:        u64, // is incremented everytime we add a valid entry
+    max_update_time_gap:        i64, // maximum time between two updates in the current entry.
```
tompntn, line 63: Could you add an explanation of the semantics of this field? At a first glance, it could be interpreted as either a limit or a measurement, and it's only when you read the implementation it becomes clear it's a measurement.

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+//running sums if needed
+use {
+    i64 as SignedTrackerRunningSum,
+    u64 as UnsignedTrackerRunningSum,
+};
+//multiply by this to make up for errors that result when dividing integers
+pub const SMA_TRACKER_PRECISION_MULTIPLIER: i64 = 10;
+
+
+pub trait Tracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
```
tompntn, line 26: These shouldn't be generic paramaters, but rather fields in a struct.

Using traits to emulate abstract classes, by implementing common behaviour in their functions, is a code smell that you're trying to use rust in an traditional OOO inheritance-style way. Things will typically be easier if you use a composition approach: a struct which has the common functionality defined on it, with a generic field for the abstract logic.

This trait should be much smaller, containing only the `add_price` method, as that is the only abstract functionality.

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+    }
+    ///returns the remining time before the next multiple of GRANUALITY
+    fn get_time_to_entry_end(time: i64) -> i64 {
+        (GRANUALITY - (time % GRANUALITY)) % GRANUALITY
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+/// Represents an SMA Tracker that has NUM_ENTRIES entries
```
tompntn, line 56: Could we use the full phrase "Simple Moving Average" here? Along with a brief description of exactly what that means in this context.

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+        Ok(())
+    }
+}
+
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+/// Represents an Tick Tracker that has NUM_ENTRIES entries
+/// each tracking the last tick before every GRANUALITY seconds.
+pub struct TickTracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
```
tompntn, line 243: Could you elaborate in this comment a bit more on what exactly "Tick Tracker" means? Maybe with an example?

program/rust/src/time_machine_types.rs
guibescos: Yep, my comment is outdated

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+        let current_entry = Self::time_to_entry(current_time)?;
+        let num_skipped_entries = current_entry - prev_entry;
+        match num_skipped_entries {
+            0 => {
+                self.prices[current_entry] = current_price;
+                self.confidences[current_entry] = current_conf;
+            }
+            1 => {
+                self.prices[prev_entry] = prev_price;
+                self.confidences[prev_entry] = prev_conf;
```
guibescos, line 295: Not sure why you need to assign here `self.prices[prev_entry]` should be `prev_price` already

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+                self.entry_validity[prev_entry] =
+                    if Self::get_time_to_entry_end(prev_time) >= THRESHOLD {
+                        0
+                    } else {
+                        1
+                    };
+                self.prices[current_entry] = current_price;
+                self.confidences[current_entry] = current_conf;
+            }
+            _ => {
```
guibescos, line 305: You can probably do the same behavior for 1 and _ in this match statement.

program/rust/src/time_machine_types.rs
guibescos: I'm almost convinced this is ok.

program/rust/src/time_machine_types.rs
guibescos: Yep, this isn't great. Let's discuss it in person

guibescos (commented):
Some comments

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,371 @@ use crate::c_oracle_header::{
...
+        Ok(num_skiped_entries)
+    }
+    ///gets the index of the next entry
+    fn get_next_entry(current_entry: usize) -> usize {
+        (current_entry + 1) % NUM_ENTRIES
+    }
+
+    ///returns the remining time before the next multiple of GRANUALITY
+    fn get_time_to_entry_end(time: i64) -> i64 {
+        (GRANUALITY - (time % GRANUALITY)) % GRANUALITY
```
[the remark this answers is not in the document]
majabbour: It is nit in case time is divisble by GRANUALity

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,366 @@ use crate::c_oracle_header::{
...
+        prev_price: i64,
+        prev_conf: u64,
+        current_time: i64,
+        current_price: i64,
+        current_conf: u64,
+    ) -> Result<(), OracleError> {
+        //multiply by precision multiplier so that our entry aggregated can be rounded to have
+        //the same accuracy as user input
+        let inflated_prev_price = prev_price * SMA_TRACKER_PRECISION_MULTIPLIER;
+        let inflated_current_price = current_price * SMA_TRACKER_PRECISION_MULTIPLIER;
```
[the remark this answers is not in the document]
majabbour: I was planning to divide by that in the sdk, the same trick is used elsewhere in our code to avoid fixed integer arithmetic errors (basically having an extra digit allows us to know in which direction to round, and since we only do one division for each entry this is good, I will document this in a better comment tomorrow)

program/rust/src/time_machine_types.rs
majabbour: i think we need to discuss what windows what we want to provide, I just left this is a placeholder

program/rust/src/time_machine_types.rs
majabbour: interesting, I think I asked Jeff about both and he seemed to agree that they are "robust", I can check with him if we want to interpolate, but that would change the validity condition.

This, Tick, and SMA, is what I recall agreeing in with atleast two other people in team before doing this. so it sounds like we need to have a discussion about what exactly we want to provide

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
 };
-use solana_program::msg;
+use std::cmp;
 
+//To make it easy to change the types to allow for more usefull
+//running sums if needed
+use {
+    i64 as SignedTrackerRunningSum,
+    u64 as UnsignedTrackerRunningSum,
+};
```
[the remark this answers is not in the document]
majabbour: sure, is there any reason why so?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+//multiply by this to make up for errors that result when dividing integers
+pub const SMA_TRACKER_PRECISION_MULTIPLIER: i64 = 10;
+
+
+pub trait Tracker<const GRANUALITY: i64, const NUM_ENTRIES: usize, const THRESHOLD: i64> {
+    ///initializes a zero initialized tracker
+    fn initialize(&mut self) -> Result<(), OracleError>;
+
+    ///add a new price to the tracker
```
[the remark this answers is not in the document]
majabbour: 1.  Why we need to pass the previous price is because of the formula of SMA tracking! once you migrate everything into one struct you won't need it, but I see no point in storing stuff that are in price_t again in each tracker. Though we can do it if you think that's is significantly cleaner

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,375 @@ use crate::c_oracle_header::{
...
+            entry = Self::get_next_entry(entry);
+        }
+    }
+    /// returns a weighted average of v with weights w
+    fn weighted_average<
+        T: std::ops::Mul<Output = T>
+            + std::ops::Add<Output = T>
+            + std::ops::Div<Output = T>
+            + std::marker::Copy,
+    >(
```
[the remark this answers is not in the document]
majabbour: I think during development, it was used on atleast two types (i64, and u64). I think if we make confidence also signed we won't need that. I just did not want to minimize switches between signed and unsigned types.

program/rust/src/time_machine_types.rs
majabbour: So it sounds like there are two orthognal comments here.
1. These const generics should not be const generics
2. there's a better way to minimize duplicate code with SMA and TickTrackers

For 1, I disagree without a clearer reason. These things are constant, one of them is litterally the size of an array. So, atleast that IMO belongs as a const genric, so please expand on 1.

For 2, this sounds right, I tried to do that in an earlier draft on my own but things got a little confusing (specially with traits not being able to access members, and the amount of boiler plate code to right getters and setters outweighting the benefits of the split). That said, this implementation has gotten much simpler as I thought more about it, so maybe, it is worth revisiting. Is there any example that you think I should look at it? I am very new to rust, certainly new to this stuff

program/rust/src/time_machine_types.rs
majabbour: Good point, I think I should look into this again. I believe it is because I was confused about when this function is being called (with two prices or only one), and wanted my code to be good either way, but ptobably this was a bad way to do it.

program/rust/src/time_machine_types.rs
majabbour: - updates to entries, 1 only updates one entry. I can do something similar to SMA trackers, and have a helper that does 1 on an entery and call it twice in _. Does that sound better?

program/rust/src/time_machine_types.rs
majabbour: Replaced with comment!

program/rust/src/time_machine_types.rs
majabbour: I agree that this is not ideal, I believe that the best way is to have a type bigger than 64 bits, and that is big enough to support our granualirities. Otherwise, there's really not much you can do beyond invalidate an entry if an overflow happens with in it. The main risk created by this is that one very bad price is enough to ruin things forever.

I can have an alternative, where instead we store the slot of the last overflow, and use that to ensure that all intermediate overflow points are included in our Wormhole sequence. Retrospectively, this seems cleaner what do you think?

Alternatively, we can choose not to support longer time averages if we believe people are unlikely to use them.

program/rust/src/time_machine_types.rs
jayantk: makes sense. fwiw you can head off these kinds of comments by putting a `// TODO: ` comment in the code. often a good trick for answering questions that you know reviewers will ask.

program/rust/src/time_machine_types.rs
jayantk: the thing I'm confused about is that `valid_entry_counter` is a purely incrementing thing. I would have expected to have a count stored per entry or something such that you can subtract?

majabbour: Oh I understand the confusion now, so, I did not see a reason to store it per entry because we can just count the number of valid entries with in the last  NUM_entries, it was primarily here for when we get two different VAAs. We can use it to infer the Number of Valid Entries before each entry.

Though storing it for each entry might make the design easier, if it is not too expensive, am happy to do either.

program/rust/src/tests/test_time_machine.txt
```diff
@@ -0,0 +1,30 @@
+test initialize for:
```
guibescos, line 1: Maybe delete this file?

majabbour: I will use it when I write tests feel free to ignore it meanwhile

majabbour: you are not gonna merge without tests anyway

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,390 @@ use crate::c_oracle_header::{
...
+        time: i64,
+        num_entries: usize,
+        granuality: i64,
+    ) -> Result<usize, OracleError> {
+        Ok(try_convert::<i64, usize>(time / granuality)? % num_entries)
+    }
+
+    ///returns the remining time before the next multiple of GRANULARITY
+    pub fn get_time_to_entry_end(time: i64, granuality: i64) -> i64 {
+        (granuality - (time % granuality)) % granuality
```
guibescos, line 56: I think this was already discussed but why is`(granuality - (time % granuality))` instead of `(granuality - (time % granuality)) % granuality` not valid?

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,390 @@ use crate::c_oracle_header::{
...
+            if self.current_entry_status == Status::Pending && slot_gap < self.threshold {
+                self.running_valid_entry_counter[prev_entry] + 1
+            } else {
+                self.running_valid_entry_counter[prev_entry]
+            };
+
+        //setup the following entry
+        self.current_entry_slot_accumelator = 0;
+        self.current_entry_status = if slot_gap < self.threshold {
+            Status::Pending
```
guibescos, line 235: I think this is handled by the subsequent call to `update_entry_price`

program/rust/src/time_machine_types.rs
majabbour: it won't produce the correct output when time is divisible by 0

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,353 @@ use crate::c_oracle_header::{
...
+
+    ///add a new price to the tracker
+    fn add_price(
+        &mut self,
+        prev_time: i64,
+        current_time: i64,
+        current_price: i64,
+        current_conf: u64,
+    ) -> Result<(), OracleError> {
+        self.close_previous_tick(prev_time, current_time)?;
```
jayantk, line 255: fwiw i think this code would be clearer if you got rid of these inner functions, as it would be like 50% of the length. (of course, there will be other times where that will be underabstracting, but this feels a little overabstracted to me)

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,353 @@ use crate::c_oracle_header::{
...
+            current_time,
+            last_two_prices,
+            last_two_confs,
+            slot_gap,
+        )?;
+        self.tick_tracker.add_price(
+            prev_time,
+            current_time,
+            last_two_prices.0,
+            last_two_confs.0,
```
jayantk, line 352: wait these tuples are ordered as `(cur_price, prev_price)`? that seems like the opposite of what someone would expect

imo if you're trying to reduce the number of arguments to these functions, a better approach would be a struct with fields like `{price, time, conf}`

jayantk: Actually, they aren't ordered that way, and this is a bug.

program/rust/src/time_machine_types.rs
```diff
@@ -35,13 +368,20 @@ pub struct PriceAccountWrapper {
...
+        self.time_machine.initialize()
     }
 
     pub fn add_price_to_time_machine(&mut self) -> Result<(), OracleError> {
-        msg!("implement me");
-        Ok(())
+        //this would work on the first add_price as well, as the first entry is always invalid, so
+        // even though we would probabpy end up overflowing the first number that should not corrupt
+        // any valid data
+        self.time_machine.add_price(
```
jayantk, line 378: I think we need to be careful to only call this function when the price is successfully updated (i.e., the current status is TRADING). do we do that in the calling code?

jayantk (commented):
this is definitely on the right track. Code clarity is much improved. I think there are some bugs here, but the only reason I can see these bugs is because the clarity has improved.

program/rust/src/time_machine_types.rs
majabbour: We only call it on upd_aggregate returns true, and it only does that when status is trading.
We can add an extra check to harden it if we want

program/rust/src/time_machine_types.rs
majabbour: Right, sorry about that. It originally was stroing the prev_tick then I swithced to this

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,353 @@ use crate::c_oracle_header::{
...
+    pub fn add_price(
+        &mut self,
+        prev_time: i64,
+        current_time: i64,
+        last_two_prices: (i64, i64),
+        last_two_confs: (u64, u64),
+        slot_gap: u64,
+    ) -> Result<(), OracleError> {
+        //multiply by precision multiplier so that our entry aggregated can be rounded to have
+        //the same accuracy as user input
```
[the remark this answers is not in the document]
majabbour: It is misplaced

program/rust/src/time_machine_types.rs
```diff
@@ -6,20 +6,353 @@ use crate::c_oracle_header::{
...
+        let prev_entry = time_to_entry(prev_time, NUM_ENTRIES, self.granularity)?;
+        let current_entry = time_to_entry(current_time, NUM_ENTRIES, self.granularity)?;
+        let num_skipped_entries = current_entry - prev_entry;
+
+        let mut entry_to_finalize = prev_entry;
+        for _ in 0..min(num_skipped_entries, NUM_ENTRIES) - 1 {
```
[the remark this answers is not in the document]
majabbour: yes

program/rust/src/time_machine_types.rs
guibescos: Can confirm this

PR #253, opened 2022-08-17
Add headers to prevent compiling errors under certain platforms.

ZhencBoris:


tompntn:
Which platforms specifically does this fix?

PR #259, opened 2022-08-19
Make it build on macs

jayantk:
the cargo build (and `./scripts/build-bpf.sh`) used were failing because we were defining these integer types multiple times. It turns out that we already have a solution to this problem in `compat_stdint.h` which  selects between the system stdint.h and the definitions in the solana sdk. 

Note that the repo still expects to be placed adjacent to a `solana` directory. I'll fix that next.

### DIFF
--- a/program/rust/build.rs
+++ b/program/rust/build.rs
@@ -9,6 +9,7 @@ fn main() {
 
     //generate and write bindings
     let bindings = Builder::default()
+        .clang_arg("-I../../../solana/sdk/bpf/c/inc/")
         .header("./src/bindings.h")
         .parse_callbacks(Box::new(parser))
         .rustfmt_bindings(true)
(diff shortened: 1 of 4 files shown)


program/rust/build.rs
```diff
@@ -9,6 +9,7 @@ fn main() {
 
     //generate and write bindings
     let bindings = Builder::default()
+        .clang_arg("-I../../../solana/sdk/bpf/c/inc/")
```
jayantk, line 12: required to make bindgen find the solana header files.

hardcoding this path is obviously bizarre. however, the C makefile already does this, so I'm not making the situation any worse.

guibescos (approved):
I'm not sure I get the changes to `build.rs` but seems minimal and it seems like you're gonna change it soon anyway.

PR #260, opened 2022-08-19
Command to delete price accounts

jayantk:
We need this command in order to clean up the situation on pythtest, where we created a bunch of incorrect product and price accounts.

This PR went down a bit of a rabbit hole, as this instruction needs to be tested with the Solana runtime. The existing Python integration tests are pretty hard to work with, so I looked for an alternative. Turns out there's a nice solana TX simulator that you can run directly from Rust, so I built some testing utils around that.


program/rust/src/time_machine_types.rs
```diff
@@ -75,7 +75,7 @@ pub mod tests {
     fn c_time_machine_size_is_correct() {
         assert_eq!(
         size_of::<TimeMachineWrapper>(),
-        TIME_MACHINE_STRUCT_SIZE.try_into().unwrap(),
+        TIME_MACHINE_STRUCT_SIZE as usize,
```
jayantk, line 78: there are two ambiguous type parameters for the `try_into` now. Something that got pulled in with tokio i think 🤷

Reisen: Looks like the right cast anyhow.

Reisen (approved):
Looks good, only a minor nitpick on the instruction de structure. The test is especially nice.

program/rust/src/rust_oracle.rs
```diff
@@ -371,6 +367,54 @@ pub fn add_price(
...
+/// price accounts for the given product.
+///
+/// Warning: This function is dangerous and will break any programs that depend on the deleted
+/// price account!
+pub fn del_price(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> ProgramResult {
+    let [funding_account, product_account, price_account] = match accounts {
```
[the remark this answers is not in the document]
jayantk: hmm yeah that does seem better. we've been doing it this way in a bunch of places already though, so i'm not going to change this right now.

guibescos (approved):

PR #262, opened 2022-08-22
Command to delete a product

jayantk:
This is again needed to clean up the pythtest situation, where we created extra product/price accounts.

### DIFF
--- a/program/rust/src/rust_oracle.rs
+++ b/program/rust/src/rust_oracle.rs
@@ -62,6 +62,7 @@ use crate::utils::{
     check_valid_writable_account,
     is_component_update,
     pubkey_assign,
+    pubkey_clear,
     pubkey_equal,
     pubkey_is_zero,
     pyth_assert,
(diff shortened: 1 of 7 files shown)


guibescos (commented):
I would add a test that deletes the last element of the array just for extra security. Otherwise LGTM.

program/rust/src/rust_oracle.rs
```diff
@@ -705,3 +706,67 @@ pub fn set_min_pub(
...
+    check_valid_signable_account(program_id, mapping_account, size_of::<pc_map_table_t>())?;
+    check_valid_signable_account(program_id, product_account, PC_PROD_ACC_SIZE as usize)?;
+
+    {
+        let cmd_args = load::<cmd_hdr_t>(instruction_data)?;
+        let mut mapping_data = load_checked::<pc_map_table_t>(mapping_account, cmd_args.ver_)?;
+        let product_data = load_checked::<pc_prod_t>(product_account, cmd_args.ver_)?;
+
+        // This assertion is just to make the subtractions below simpler
+        pyth_assert(mapping_data.num_ >= 1, ProgramError::InvalidArgument)?;
```
Reisen, line 737: Rather than assert, can we used `.checked_sub` on the subtractions below? We can (and should) enable clippy’s arithmetic lint which will be safer. We should really be using checked arithmetic for everything on-chain.

jayantk: good call on the arithmetic checks. will enable in a separate pr.

PR #263, opened 2022-08-22
Move compilation warnings to CI

jayantk:
`#![deny(warnings)]` reduces development speed because you have to fix all this stuff in order to build/run tests. This makes it harder to iterate on bug fixes, e.g., to quickly change something in your code to see if it fixes a test. It's also bad practice for rust because it can break the build when something in rust changes (see https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html  ).

This PR moves the warning check to CI.


guibescos:
I think you got rid of the formatting checks.

PR #264, opened 2022-08-22
Rust instructions

guibescos:
First step to delete the instruction enum from `oracle.h`
- `load` and `load_mut` now fail with custom errors.
- Rust enum for the dispatcher with a way to cast from the command header to the rust enum.


program/rust/src/processor.rs
```diff
@@ -45,36 +29,26 @@ pub fn process_instruction(
...
+        OracleCommand::UpdTest => Err(OracleError::UnrecognizedInstruction.into()),
+        OracleCommand::SetMinPub => set_min_pub(program_id, accounts, instruction_data),
+        OracleCommand::UpdPriceNoFailOnError => {
             upd_price_no_fail_on_error(program_id, accounts, instruction_data)
         }
-        command_t_e_cmd_resize_price_account => {
+        OracleCommand::ResizePriceAccount => {
             resize_price_account(program_id, accounts, instruction_data)
         }
-        command_t_e_cmd_add_price => add_price(program_id, accounts, instruction_data),
```
jayantk, line 68: can you delete the enum in `oracle.h` so no one can depend on it in the C?

program/rust/src/instruction.rs
```diff
@@ -0,0 +1,101 @@
...
+    // account[0] funding account       [signer writable]
+    // account[1] price account         [signer writable]
+    // account[2] system program        []
+    ResizePriceAccount,
+    // deletes a price account
+    // account[0] funding account       [signer writable]
+    // account[1] product account       [signer writable]
+    // account[2] price account         [signer writable]
+    DelPrice,
+}
```
jayantk, line 85: I suggest adding a test for OracleCommand that tests the `i32` value of each entry in the enum. We've accidentally added elements to the middle of this list before, which is a non-backward compatible change.

Also please document that commands *must* be added to the end of this list

jayantk (commented):
generally fine. couple minor comments

program/rust/src/processor.rs
guibescos: Some things still depend, just wanted to stage it incrementally

program/rust/src/instruction.rs
guibescos: Done, I explicitly wrote the variants.

PR #265, opened 2022-08-22
Cleanup rust oracle

guibescos:
Updating `rust_oracle.rs` to use the new structs.

There's an annoying Pod derive failure with SetMinPubArgs because it has padding. Accessing the padding bytes is UB I think.


program/rust/src/instruction.rs
```diff
@@ -105,3 +108,44 @@ pub fn load_command_header_checked(data: &[u8]) -> Result<OracleCommand, OracleE
...
+    pub pub_:   pc_pub_key_t,
+}
+
+pub type DelPublisherArgs = AddPublisherArgs;
+
+
+unsafe impl Pod for SetMinPubArgs {
+}
+/// This is embarassing, Pod derivation doesn't work for this struct because it has padding in the
+/// end WARNING: DO NOT ACCESS THE 3 LAST BYTES OF THIS STRUCT!
```
jayantk, line 134: last 2 bytes right?

also how about adding `pub unused_: u8` as the last field to get the Pod derivation to work?

program/rust/src/instruction.rs
```diff
@@ -105,3 +108,44 @@ pub fn load_command_header_checked(data: &[u8]) -> Result<OracleCommand, OracleE
     }
     OracleCommand::from_i32(command_header.command).ok_or(OracleError::UnrecognizedInstruction)
 }
+
+#[repr(C)]
+#[derive(Zeroable, Pod, Copy, Clone)]
+pub struct AddPriceArgs {
+    pub header: CommandHeader,
+    pub expo_:  i32,
```
jayantk, line 116: why the _ suffix on these fields?

jayantk (commented):
cool, just a couple minor comments.

program/rust/src/instruction.rs
guibescos: You're so right!

program/rust/src/instruction.rs
guibescos: Fixed

PR #267, opened 2022-08-23
Rust structs

guibescos:
Oracle structs in rust.


program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,128 @@ pub trait PythAccount: Pod {
...
-
-#[cfg(target_endian = "little")]
-unsafe impl Pod for pc_price_comp_t {
-}
+    const INITIAL_SIZE: u32 = 240; // Tested in test_sizes.rs
+}
 
-impl pc_pub_key_t {
-    pub fn new_unique() -> pc_pub_key_t {
+#[repr(C)]
```
guibescos, line 55: Trying to not change names too much because some of the fields are used in `upd_aggregate.c`

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,128 @@ pub trait PythAccount: Pod {
     }
 }
 
-impl PythAccount for pc_map_table_t {
+impl PythAccount for MappingAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_MAPPING;
-    const INITIAL_SIZE: u32 = PC_MAP_TABLE_T_PROD_OFFSET as u32;
+    const INITIAL_SIZE: u32 = 56; // Tested in test_sizes.rs
```
guibescos, line 39: I'm trying to avoid `as` everywhere. It's tricky to use `try_from` here because it's a constant. The solution is hardcoding and testing this in `test_size.rs`.

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,128 @@ pub trait PythAccount: Pod {
...
+#[derive(Copy, Clone)]
+pub struct MappingAccount {
+    pub header:               AccountHeader,
+    pub number_of_products:   u32,
+    pub unused_:              u32,
+    pub next_mapping_account: CPubkey,
+    pub product_list:         [CPubkey; 640usize],
+}
+
+// Unsafe impl because pc_pub_key is a union
```
guibescos: We will be able to fix this if we switch to a non-union implementation of pubkey.

program/rust/src/time_machine_types.rs
```diff
@@ -12,6 +11,8 @@ use bytemuck::{
 };
 use solana_program::msg;
 
+#[allow(dead_code)]
```
tompntn, line 14: Curious why this is necessary?

program/rust/src/c_oracle_header.rs
tompntn: Could you add a comment explaining how this value is calculated?

program/rust/src/c_oracle_header.rs
tompntn: I don't think the `k1_` variant is ever used - I think we can just use the `k8_` variant and avoid the use of unsafe.

guibescos: Both variants are used.

program/rust/src/time_machine_types.rs
guibescos: Not doing this anymore in the new PR.

PR #270, opened 2022-08-24
 Create rust structs and replace bindgen structs

guibescos:
This PR :
- Creates rust structs in `c_oracle_header.rs` and replaces all the bindgen structs by the new structs.

### DIFF
--- a/program/rust/src/c_oracle_header.rs
+++ b/program/rust/src/c_oracle_header.rs
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
     }
 }
 
-impl PythAccount for pc_map_table_t {
+impl PythAccount for MappingAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_MAPPING;
     const INITIAL_SIZE: u32 = PC_MAP_TABLE_T_PROD_OFFSET as u32;
 }
 
-impl PythAccount for pc_prod_t {
+impl PythAccount for ProductAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRODUCT;
-    const INITIAL_SIZE: u32 = size_of::<pc_prod_t>() as u32;
+    const INITIAL_SIZE: u32 = size_of::<ProductAccount>() as u32;
     fn minimum_size() -> usize {
         PC_PROD_ACC_SIZE as usize
     }
 }
 
-impl PythAccount for pc_price_t {
+impl PythAccount for PriceAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRICE;
     const INITIAL_SIZE: u32 = PC_PRICE_T_COMP_OFFSET as u32;
 }
 
-#[cfg(target_endian = "little")]
-unsafe impl Zeroable for pc_acc {
+#[repr(C)]
+#[derive(Copy, Clone, Pod, Zeroable)]
+pub struct PriceAccount {
+    pub magic_:          u32,
+    pub ver_:            u32,
+    pub type_:           u32,
+    pub size_:           u32,
+    pub ptype_:          u32,      // Type of the price account
+    pub expo_:           i32,      // Exponent for the published prices
+    pub num_:            u32,      // Current number of authorized publishers
+    pub num_qt_:         u32,      // Number of valid quotes for the last aggregation
+    pub last_slot_:      u64,      // Last slot with a succesful aggregation (status : TRADING)
+    pub valid_slot_:     u64,      // Second to last slot where aggregation was attempted
+    pub twap_:           PriceEma, // Ema for price
+    pub twac_:           PriceEma, // Ema for confidence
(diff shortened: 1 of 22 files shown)


program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
...
+
+// Unsafe impl because product_list is of size 640 and there's no derived trait for this size
+unsafe impl Pod for MappingAccount {
+}
+unsafe impl Zeroable for MappingAccount {
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union CPubkey {
```
guibescos, line 159: Both fields are currently used in the code. I feel like we can get to a safe derivation if we :
- Replace CPubkey by the solana sdk Pubkey 
- Update functions to only use `k1_` or `k8_`.
Let's do it in a future PR.

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
...
-
-impl pc_pub_key_t {
-    pub fn new_unique() -> pc_pub_key_t {
+#[repr(C)]
+#[derive(Copy, Clone, Pod, Zeroable)]
+pub struct PriceAccount {
+    pub magic_:          u32,
+    pub ver_:            u32,
+    pub type_:           u32,
+    pub size_:           u32,
```
Reisen, line 61: Can we move these in-line comments above the fields and make them Rust doc comments:

```suggestion
    /// Example documentation comment with a triple ///
    pub size_:           u32,
```

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
...
+    pub unused_3_:       i32,
+    pub prod_:           CPubkey, // Corresponding mapping account
+    pub next_:           CPubkey, // Next price account in the list
+    pub prev_slot_:      u64,     /* Second to last slot where aggregation was succesful
+                                   * (status : TRADING) */
+    pub prev_price_:     i64,       // Aggregate price at prev_slot_
+    pub prev_conf_:      u64,       // Confidence interval at prev_slot_
+    pub prev_timestamp_: i64,       // Timestamp of prev_slot_
+    pub agg_:            PriceInfo, // Last attempted aggregate results
+    pub comp_:           [PriceComponent; 32usize], // Publisher's price components
```
Reisen, line 83: Nitpick: Doesn’t need fixing but array sizes are always `usize` so can be specified as just 32.

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
...
+#[derive(Copy, Clone)]
+pub struct MappingAccount {
+    pub magic_:  u32,
+    pub ver_:    u32,
+    pub type_:   u32,
+    pub size_:   u32,
+    pub num_:    u32,
+    pub unused_: u32,
+    pub next_:   CPubkey,
+    pub prod_:   [CPubkey; 640usize],
```
Reisen, line 141: Can we lift the 640usize and others to a `const`, similar to the C headers. It’s easier to spot file-level constants as they show up as language items in IDEs.

program/rust/src/c_oracle_header.rs
```diff
@@ -34,94 +34,137 @@ pub trait PythAccount: Pod {
     }
 }
 
-impl PythAccount for pc_map_table_t {
+impl PythAccount for MappingAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_MAPPING;
     const INITIAL_SIZE: u32 = PC_MAP_TABLE_T_PROD_OFFSET as u32;
```
Reisen, line 39: Can we not use `size_of::<MappingAccount>` here as in `ProductAccount`?

program/rust/src/c_oracle_header.rs
Reisen: Agree with the first suggestion, we should use the solana `Pubkey` type directly in a future PR. The second suggestion would just keep us using the ad-hoc unsafe functions we should scrap.

program/rust/src/c_oracle_header.rs
guibescos: In the future they'll be constants

program/rust/src/c_oracle_header.rs
guibescos: No, added a comment explaning.

program/rust/src/c_oracle_header.rs
guibescos: Yes, I agree. Next PR will be getting the constants from oracle.h . As of now, I'll use the C constant for this PR.

program/rust/src/c_oracle_header.rs
guibescos: Good

Reisen (approved):

PR #271, opened 2022-08-24
Move minimum size check to load_checked and initialize_checked

guibescos:
- Move the `minimum_size`  check to `load_checked` and `initialize_checked` from `check_valid_signable_account` and `check_valid_writable_account`. This allows to leverage the PythAccount trait for the check. It makes sense to have that check in the deserialization functions because if the account is too small deserialization will fail.
 - Move `check_valid_fresh_account` to `initialize_pyth_account_checked`. These two functions are always called together. Makes it harder to accidentally reinitialize an account.

### DIFF
--- a/program/rust/src/deserialize.rs
+++ b/program/rust/src/deserialize.rs
@@ -13,6 +13,7 @@ use crate::c_oracle_header::{
 };
 use crate::error::OracleError;
 use crate::utils::{
+    check_valid_fresh_account,
     clear_account,
     pyth_assert,
 };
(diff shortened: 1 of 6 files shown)


program/rust/src/deserialize.rs
```diff
@@ -46,6 +48,7 @@ pub fn load_mut<T: Pod>(data: &mut [u8]) -> Result<&mut T, OracleError> {
 }
 
 /// Get the data stored in `account` as a value of type `T`
+/// /// Warning: DON'T USE THIS UNCHECKED FUNCTION
 pub fn load_account_as<'a, T: Pod>(account: &'a AccountInfo) -> Result<Ref<'a, T>, ProgramError> {
     let data = account.try_borrow_data()?;
```
guibescos, line 53: I want to make this private but it'll require some more refactoring.

jayantk (approved):

PR #273, opened 2022-08-25
Add test for resize instruction

guibescos:
Add a new test for the resize instruction. This instruction will enable price account resizing for adding more publishers and the settlement prices.


jayantk (commented):
would be good to test the permissions on this instruction also, but that can happen later.

jayantk (approved):

PR #275, opened 2022-08-28
Init permissions

guibescos:
Adds an instruction called `upd_permissions` which initializes or updates the `PermissionAccount`.
This instructions needs to get called by the update authority of the program. In order to check that the signer is the actual update authority, we need to pass the `program` account and the `program_data` account as well.


program/rust/src/tests/pyth_simulator.rs
```diff
@@ -65,9 +80,92 @@ impl PythSimulator {
...
             last_blockhash: recent_blockhash,
+            programdata_id: None,
         }
     }
 
+    /// Deploys the oracle program as upgradable
+    pub async fn new_upgradable() -> PythSimulator {
+        let mut bpf_data = read_file(find_file("pyth_oracle.so").unwrap_or_else(|| {
+            panic!("Unable to locate {}", "pyth_oracle.so");
+        }));
```
jayantk, line 91: is the build / test process guaranteed to produce this `.so` file before the test runs? That is, is there any possibility of running this test with an outdated binary?

program/rust/src/tests/pyth_simulator.rs
```diff
@@ -65,9 +80,92 @@ impl PythSimulator {
...
+
+        // Same as new() but here we set the payer to be upgrade_authority and program_id to be the
+        // upgradable program
+        return PythSimulator {
+            program_id: program_key,
+            banks_client,
+            payer: upgrade_authority_keypair,
+            last_blockhash: recent_blockhash,
+            programdata_id: Some(programdata_key),
+        };
```
jayantk, line 165: i would be in favor of replacing `new` with this method so there's only one simulator setup for tests. I also suggest adding `upgrade_authority` as a separate field to the simulator struct so that users can test with an arbitrary key and the upgrade authority.

program/rust/src/tests/test_upd_permissions.rs
```diff
@@ -0,0 +1,89 @@
...
+    //Check that the program got deployed properly and can be interacted with
+    let mapping_keypair = sim.init_mapping().await.unwrap();
+    let product = sim.add_product(&mapping_keypair).await.unwrap();
+    let price = sim.add_price(&product, -8).await.unwrap();
+    assert!(sim.get_account(price.pubkey()).await.is_some());
+
+    let mut master_authority = CPubkey::new_unique();
+    let mut data_curation_authority = CPubkey::new_unique();
+    let mut security_authority = CPubkey::new_unique();
+
```
jayantk, line 28: I suggest adding some permissions tests here as well to ensure e.g., that only the upgrade authority can invoke this instruction.

I don't know the best way to set up those tests, but it's worth thinking about because we'll definitely need the same tests for all of the other instructions going forward.

jayantk (commented):
this seems reasonable to me, but I'd like @Reisen to approve since there's some tricky solana stuff here.

program/rust/src/error.rs
```diff
@@ -7,31 +7,37 @@ use thiserror::Error;
...
-    InvalidInstructionVersion      = 609,
+    InvalidInstructionVersion            = 609,
     #[error("InstructionDataTooShort")]
-    InstructionDataTooShort        = 610,
+    InstructionDataTooShort              = 610,
     #[error("InstructionDataSliceMisaligned")]
-    InstructionDataSliceMisaligned = 611,
+    InstructionDataSliceMisaligned       = 611,
+    #[error("DeserializingError")]
+    DeserializingError                   = 612,
```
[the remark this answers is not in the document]
guibescos: Done

program/rust/src/error.rs
```diff
@@ -7,31 +7,37 @@ use thiserror::Error;
...
+    #[error("FailedAuthenticatingUpgradeAuthority")]
+    FailedAuthenticatingUpgradeAuthority = 613,
+    #[error("FailedPDAVerification")]
+    FailedPDAVerification                = 614,
```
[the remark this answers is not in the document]
guibescos: Done

program/rust/src/tests/pyth_simulator.rs
guibescos: Yes `cargo test-bpf` generates this if the source code has changed.

program/rust/src/tests/pyth_simulator.rs
guibescos: Good

program/rust/src/tests/test_upd_permissions.rs
guibescos: Made some helpers for changing the payer

program/rust/src/tests/pyth_simulator.rs
```diff
@@ -34,41 +41,125 @@ use crate::c_oracle_header::{
...
+            .await
+            .unwrap();
+
+        recent_blockhash = blockhash;
+
+        transaction.partial_sign(&[&payer], recent_blockhash);
+
+        banks_client.process_transaction(transaction).await.unwrap();
+
+        let genesis_keypair = Keypair::from_bytes(&payer.to_bytes()).unwrap(); // Hack to duplicate a keypair
```
guibescos, line 141: No very happy about this. I think Keypair doesn't have a Copy trait because solana doesn't want people to copy keypairs.

program/rust/src/tests/test_upd_permissions.rs
guibescos: And added a test

program/rust/src/error.rs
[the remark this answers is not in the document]
tompntn: The Rust naming conventions also suggest treating acronoyms as single words: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

program/rust/src/c_oracle_header.rs
```diff
@@ -54,6 +56,22 @@ impl PythAccount for PriceAccount {
...
+impl PythAccount for PermissionAccount {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PERMISSIONS;
+    const INITIAL_SIZE: u32 = size_of::<PermissionAccount>() as u32;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Pod, Zeroable)]
+pub struct PermissionAccount {
+    pub header:                  AccountHeader,
+    pub bump:                    u8,
```
tompntn, line 68:
```suggestion
    pub bump_seed:               u8,
```

program/rust/src/rust_oracle.rs
```diff
@@ -803,3 +836,77 @@ pub fn del_product(
 
     Ok(())
 }
+
+
+pub fn upd_permissions(
```
tompntn, line 841: Could you add a comment explaining the semantics of `upd_permissions`, like there are on the other public functions (e.g. `del_product`)? In general docstrings on top-level public functions are very valuable.

program/rust/src/error.rs
```diff
@@ -7,31 +7,37 @@ use thiserror::Error;
...
+    InstructionDataTooShort              = 610,
     #[error("InstructionDataSliceMisaligned")]
-    InstructionDataSliceMisaligned = 611,
+    InstructionDataSliceMisaligned       = 611,
+    #[error("DeserializationError")]
+    DeserializationError                 = 612,
+    #[error("FailedAuthenticatingUpgradeAuthority")]
+    FailedAuthenticatingUpgradeAuthority = 613,
+    #[error("FailedPdaVerification")]
+    FailedPdaVerification                = 614,
```
tompntn, line 40: `InvalidUpgradeAuthority` and `InvalidPda` would be better names: "failed" refers to the consequence of the error being thrown, not the error itself.

program/rust/src/rust_oracle.rs
```diff
@@ -803,3 +836,77 @@ pub fn del_product(
...
+        program_account,
+        programdata_account,
+        program_id,
+    )?;
+    let bump = check_pda(
+        permissions_account,
+        &[PERMISSIONS_SEED.as_bytes()],
+        program_id,
+    )?;
+    pyth_assert(
```
tompntn, line 866: `check_pda` should not return a `bump` - could we move the `find_program_address` into the top-level `upd_permissions` function, so that `check_pda` just becomes a thin wrapper around the `pyth_assert`, like the other `check_*` functions?

program/rust/src/rust_oracle.rs
```diff
@@ -803,3 +836,77 @@ pub fn del_product(
...
+        check_id(system_program.key),
+        OracleError::InvalidSystemAccount.into(),
+    )?;
+
+
+    // Create account if it doesn't exist
```
tompntn, line 872:
```suggestion
    // Create permissions account if it doesn't exist
```

program/rust/src/rust_oracle.rs
```diff
@@ -803,3 +836,77 @@ pub fn del_product(
...
+            PermissionAccount::minimum_size(),
+            program_id,
+            &[PERMISSIONS_SEED.as_bytes(), &[bump]],
+        )?;
+        check_valid_fresh_account(permissions_account)?;
+        initialize_pyth_account_checked::<PermissionAccount>(
+            permissions_account,
+            cmd_args.header.version,
+        )?;
+    }
```
tompntn, line 887: This can be factored out into a helper function.

program/rust/src/error.rs
guibescos: Yes

program/rust/src/rust_oracle.rs
guibescos: Yes

program/rust/src/rust_oracle.rs
guibescos: Done

program/rust/src/c_oracle_header.rs
```diff
@@ -54,6 +56,22 @@ impl PythAccount for PriceAccount {
...
+    pub unused_:                 [u8; 7],
```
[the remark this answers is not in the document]
Reisen: Even with the added comment, I’m not sure I understand the memory safety issue or why this field needs to be added by hand.

program/rust/src/c_oracle_header.rs
Reisen: What’s the reason we’re storing the bump seed?

program/rust/src/deserialize.rs
```diff
@@ -112,3 +117,45 @@ pub fn initialize_pyth_account_checked<'a, T: PythAccount>(
 
     load_account_as_mut::<T>(account)
 }
+
+fn create_pda_account<'a>(
```
Reisen, line 121: This way of creating accounts isn’t safe for PDA’s, as a malicious actor can calculate them upfront and deposit a single lamport in each to make the accounts un-useable.

This is because the create_account instruction refuses to create an account with non-zero lamports, which is also being done further down in `create_pda_if_needed`.

Instead, [see how solitaire create’s all accounts, which is also similar to how Anchor works.](https://github.com/wormhole-foundation/wormhole/blob/dev.v2/solana/solitaire/program/src/types/accounts.rs#L170)

program/rust/src/c_oracle_header.rs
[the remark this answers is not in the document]
guibescos: Deleted bump_seed so now I can do a #repr(C)

program/rust/src/c_oracle_header.rs
guibescos: Small efficiency gain when verifying PDAs. Although I ended up deciding against it.

program/rust/src/deserialize.rs
guibescos: Thank you sir!

program/rust/src/c_oracle_header.rs
```diff
@@ -58,6 +60,35 @@ impl PythAccount for PriceAccount {
...
+    /// - Add mapping accounts
+    /// - Add price accounts
+    /// - Add product accounts
+    /// - Delete price accounts
+    /// - Delete product accounts
+    /// - Update product accounts
+    pub data_curation_authority: Pubkey,
+    /// An authority that can  :
+    /// - Add publishers
+    /// - Delete publishers
```
jayantk, line 87: (not a comment for now) I think we should think harder about the exact permissions given to these keys. I think add/remove/min publishers should belong to the data curation authority, because that is part of curating a data feed. The security authority should get the ability to temporarily depermission a small number of publishers or something ilke that.

program/rust/src/deserialize.rs
```diff
@@ -112,3 +117,29 @@ pub fn initialize_pyth_account_checked<'a, T: PythAccount>(
...
+// Creates pda if needed and initializes it as one of the Pyth accounts
+pub fn create_pda_if_needed<'a, T: PythAccount>(
+    account: &AccountInfo<'a>,
+    funding_account: &AccountInfo<'a>,
+    system_program: &AccountInfo<'a>,
+    program_id: &Pubkey,
+    seeds: &[&[u8]],
+    version: u32,
+) -> Result<(), ProgramError> {
+    let target_rent = Rent::default().minimum_balance(T::MINIMUM_SIZE);
```
jayantk, line 130: you have to use `Rent::get()` not `Rent::default()` because we changed the minimum rent on Pythnet. (We just ran into this problem in the attester code)

program/rust/src/tests/pyth_simulator.rs
```diff
@@ -34,41 +41,112 @@ use crate::c_oracle_header::{
...
+        return result;
+    }
+
+
+    pub fn set_generic_as_payer(&mut self) {
+        self.payer = Keypair::from_bytes(&self.genesis_keypair.to_bytes()).unwrap();
+    }
+
+    pub fn set_upgrade_authority_as_payer(&mut self) {
+        self.payer = Keypair::from_bytes(&self.upgrade_authority.to_bytes()).unwrap();
```
jayantk, line 149: I suggest passing `payer` as an argument to `upd_permissions` instead of doing this. it'll be easier to understand and it's more flexible.

program/rust/src/utils.rs
```diff
@@ -135,3 +145,102 @@ pub fn is_component_update(cmd_args: &UpdPriceArgs) -> Result<bool, OracleError>
...
 }
+
+
+/// These 3 accounts need to get passed to make sure that the upgrade authority is signing the
+/// transaction
+/// - `program_account` is the program at address `program_id`. It just contains a pointer to the
+///   `programdata_account`
+/// - `programdata_account` has an `upgrade_authority_address` field that needs to match
+///   `upgrade_authority.key`
+pub fn check_is_valid_upgradeauthority_program_programdata_triplet(
```
jayantk, line 156: typo: upgradeauthority

jayantk: you could also call this `check_is_upgrade_authority_for_program` . (I realize it does a bit more than that, but all of those checks are means to the end of checking the upgrade authority.)

jayantk (commented):
lgtm aside from a couple minor things. I'll let @Reisen approve though

program/rust/src/deserialize.rs
guibescos: Can we do this in a future PR?

program/rust/src/tests/pyth_simulator.rs
guibescos: Meh the reason I ended up doing this is because I liked everything being private inside of pyth_simulator

program/rust/src/utils.rs
guibescos: YEs

program/rust/src/tests/pyth_simulator.rs
guibescos: Done

Reisen (approved):

PR #276, opened 2022-08-29
Renames

guibescos:
Renaming fields in structs now that they're in Rust. Not touching most of the structs from `pc_price_t` (`PriceAccount`) because they're still used by upd_aggregate.


jayantk (approved):
🙇 huge readability improvement here

PR #278, opened 2022-08-30
Update example blockdaemon nginx pythnet config

tompntn:
Updating the example blockdaemon nginx config to use separate pythnet servers.

### DIFF
--- a/doc/example-nginx-blockdaemon.conf
+++ b/doc/example-nginx-blockdaemon.conf
@@ -14,31 +14,49 @@ http {
 
   client_max_body_size 0;
 
-  upstream bdnodes {
-    server YOUR_NODE.bdnodes.net:443;
-    server YOUR_NODE.bdnodes.net:443;
-    server YOUR_NODE.bdnodes.net:443;
-    keepalive 4;
-  }
-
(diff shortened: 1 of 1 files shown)


doc/example-nginx-blockdaemon.conf
```diff
@@ -14,31 +14,49 @@ http {
...
+  }
+  server {
+    listen 7899;
+    server_name pythnet.localhost;
+    location / {
+      proxy_http_version 1.1;
+      proxy_set_header Connection "";
+      proxy_ssl_server_name on;
+      proxy_set_header Host "rpc.mainnet.pythnet.blockdaemon.tech";
+      proxy_pass https://rpc.mainnet.pythnet.blockdaemon.tech:443/;
```
jayantk, line 37: there's no token auth on the pythnet blockdaemon rpc nodes? (I presume this is their public RPC endpoint?)

jayantk (approved):

doc/example-nginx-blockdaemon.conf
tompntn: The Pythnet endpoints are public.

PR #285, opened 2022-09-02
Sma

guibescos:


program/rust/src/time_machine_types.rs
```diff
@@ -3,56 +3,147 @@ use crate::c_oracle_header::{
...
+        let epoch_0 = self.time_to_epoch(datapoint.last_two_timestamps.0)?;
+        let epoch_1 = self.time_to_epoch(datapoint.last_two_timestamps.1)?;
+
+        let datapoint_numerator = i128::from(datapoint.slot_gap) * i128::from(datapoint.price); //Can't overflow because u64::MAX * i64::MAX = i28::MAX
+        let datapoint_denominator = datapoint.slot_gap;
+
+        // Can't overflow, it's always smaller than the current solana slot
+        self.current_epoch_denominator += datapoint_denominator;
+        // Can't overflow, it's always smaller than u64::MAX * i64::MAX = i28::MAX
+        self.current_epoch_numerator += datapoint_numerator;
```
jayantk, line 88: not sure the overflow comment is true. `datapoint_numerator` could be `== i128::MAX` right? If I add up a few data points into `current_epoch_numerator`, then I think this can overflow?

program/rust/src/time_machine_types.rs
```diff
@@ -3,56 +3,147 @@ use crate::c_oracle_header::{
...
+    }
+
+    pub fn conclude_epoch_and_initialize_next(
+        &mut self,
+        epoch: usize,
+        datapoint_numerator: i128,
+        datapoint_denominator: u64,
+    ) {
+        let index = epoch.rem_euclid(NUM_ENTRIES);
+        let prev_index = (epoch + NUM_ENTRIES - 1).rem_euclid(NUM_ENTRIES);
```
jayantk, line 109: isn't this equal to `(epoch - 1).rem_euclid(NUM_ENTRIES)`?

jayantk: oh is this because `usize` doesn't like negative numbers?

program/rust/src/time_machine_types.rs
```diff
@@ -3,56 +3,147 @@ use crate::c_oracle_header::{
...
+        // epochs
+        if self.current_epoch_is_valid {
+            self.running_valid_epoch_counter[index] =
+                self.running_valid_epoch_counter[prev_index] + 1;
+        } else {
+            self.running_valid_epoch_counter[index] = self.running_valid_epoch_counter[prev_index]
+        };
+
+        self.current_epoch_denominator = datapoint_denominator;
+        self.current_epoch_numerator = datapoint_numerator;
```
jayantk, line 126: i'm not sure i follow this update. I would have expected both of these to get set to zero. I think this is double (really multiple) counting updates that span epoch boundaries?

jayantk (commented):
this generally seems reasonable . i left you a couple questions/comments on the implementation. I didn't read the tests carefully either.

program/rust/src/time_machine_types.rs
guibescos: Added a comment with better math. Basically :
`current_epoch_numerator = slot_gap1 * price1 + slot_gap2 * price2 + slot_gap3 * price3 <= (slot_gap1 + slot_gap2 + slot_gap3) * i64::MAX <= u64::MAX * i64::MAX

program/rust/src/time_machine_types.rs
guibescos: Yep `usize` is unsigned so this works better with tests.

program/rust/src/time_machine_types.rs
guibescos: Talked about in the call.

PR #288, opened 2022-09-07
Sma3

guibescos:


program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
-    //TimeMachine
-    pub time_machine:          TimeMachineWrapper,
+    // TimeMachine
+    pub time_machine:          SmaTracker<NUM_BUCKETS_THIRTY_MIN>,
+}
+
+
+#[derive(Debug)]
+pub struct DataPoint {
+    pub last_two_timestamps: (i64, i64),
```
jayantk, line 34: nitpick: i would prefer two separate fields here, cur_timestamp and prev_timestamp. The tuple doesn't make it clear which element of the tuple is the current/previous.

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
         Ok(())
     }
 
     pub fn add_price_to_time_machine(&mut self) -> Result<(), OracleError> {
-        // TO DO
+        #[cfg(test)]
+        self.time_machine.add_datapoint( &DataPoint{
+            last_two_timestamps : (self.price_data.prev_timestamp_, self.price_data.timestamp_),
+            slot_gap : (self.price_data.last_slot_ - self.price_data.prev_slot_),
+            price: self.price_data.prev_price_ /2  + self.price_data.agg_.price_ / 2 + (self.price_data.prev_price_ % 2) * (self.price_data.agg_.price_ % 2), // Hack to avoid overflow?
```
jayantk, line 52: why the ? at the end of the comment here

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+        }
+        )?;
+        Ok(())
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct SmaTracker<const NUM_BUCKETS: usize> {
+    /// The maximum gap in slots for an epoch's sma to be valid, we set it to PC_MAX_SEND_LATENCY
```
jayantk, line 62: get rid of "we set it to...". no reason that this needs to be specified in a code comment, because someone can just find where the struct is initialized.

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+    pub threshold:                     u64,
+    /// The legth of one sma epoch.
```
jayantk, line 64: length

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+    pub granularity:                   i64,
+    /// Numerator for the current epoch (ex : slot_gap1 * price_1 + slot_gap2 * price_2 + slot_gap3
+    /// * price_3)
+    /// It can never overflow because :
+    /// slot_gap1 * price_1 + slot_gap2 * price_2 + slot_gap3 * price_3 <= (slot_gap1 + slot_gap2 +
+    /// slot_gap3) * i64::MAX <= clock_slot * i128::MAX
```
jayantk, line 70: i think this is supposed to say `clock_slot * i64::MAX <= i128::MAX`

jayantk: 👍  for the detailed comments explaining the overflow conditions.

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+    pub current_epoch_numerator:       i128,
+    /// Denominator for the current epoch (ex : slot_gap1 + slot_gap2 + slot_gap3)
+    /// It can never overflow because : (slot_gap1 + slot_gap2 + slot_gap3) <= clock_slot
+    pub current_epoch_denominator:     u64,
+    /// Whether the current epoch is valid (i.e. max slot_gap <= threshold) for all the epochs
+    /// updates
+    pub current_epoch_is_valid:        bool,
+    /// Stores the running sum of individual epoch averages, we can support u64::MAX epochs. And
```
jayantk, line 78: clarify that `average = current_epoch_numerator / current_epoch_denominator` ?

also I don't get the part here about supporting `u64::MAX` epochs. This is because epochs are indexed by a `u64`?

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+impl<const NUM_BUCKETS: usize> SmaTracker<NUM_BUCKETS> {
+    pub fn time_to_epoch(&self, time: i64) -> Result<usize, OracleError> {
+        try_convert::<i64, usize>(time / self.granularity) // Can never fail because usize is u64
+                                                           // and time is positive
+    }
+
+    pub fn initialize(&mut self, granularity: i64, threshold: u64) {
+        self.threshold = threshold;
+        self.granularity = granularity;
+        self.current_epoch_is_valid = true;
```
jayantk, line 95: shouldn't this start as `false` because we're initializing in the middle of an epoch?

program/rust/src/time_machine_types.rs
jayantk: oh, is this because the content of this is an average price (so at most i64::MAX) added up for `n` epochs, so `u64::MAX * i64::MAX`? If so, clarify that in the comment.

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+        if epoch_0 < epoch_1 {
+            let index = epoch_0.rem_euclid(NUM_BUCKETS);
+
+            // This addition can only overflow if NUM_BUCKETS ~ u64::MAX / 2
+            let prev_index = (epoch_0 + NUM_BUCKETS - 1).rem_euclid(NUM_BUCKETS);
+
+            self.running_sum_of_price_averages[index] = self.running_sum_of_price_averages
+                [prev_index]
+                + self.current_epoch_numerator / i128::from(self.current_epoch_denominator);
+            // The fraction here is smaller than i64::MAX , so we can support u64::MAX updates
```
jayantk, line 121: weird location for this comment. put it above the line of code that it applies to, not below.

program/rust/src/tests/test_sma.rs
```diff
@@ -0,0 +1,168 @@
...
+                                                        * has been a succesful aggregation */
+            price:    i64::arbitrary(g),
+        }
+    }
+}
+
+
+#[quickcheck]
+fn test_sma(input: Vec<DataEvent>) -> bool {
+    // No gaps, no skipped epochs
```
jayantk, line 31: suggest leaving a comment above this explaining how the tests are set up, specifically that the timegap / slot_gap can be at most `u8::MAX`. It took me a bit to understand these comments about skips / gaps.

program/rust/src/tests/test_sma.rs
```diff
@@ -0,0 +1,168 @@
...
+struct DataEvent {
+    timegap:  i64,
+    slot_gap: u64,
+    price:    i64,
+}
+
+impl Arbitrary for DataEvent {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        DataEvent {
+            timegap:  i64::from(u8::arbitrary(g)),
```
jayantk, line 20: time_gap for consistency

program/rust/src/tests/test_sma.rs
```diff
@@ -0,0 +1,168 @@
+use quickcheck::Arbitrary;
```
jayantk, line 1: 🥇 for this generative test. This gives me a lot of confidence in the implementation.

program/rust/src/time_machine_types.rs
```diff
@@ -1,45 +1,192 @@
...
+pub struct DataPoint {
+    pub last_two_timestamps: (i64, i64),
+    pub slot_gap:            u64,
+    pub price:               i64,
 }
+
 impl PriceAccountWrapper {
     pub fn initialize_time_machine(&mut self) -> Result<(), OracleError> {
-        // TO DO
+        #[cfg(test)]
```
jayantk, line 41: maybe comment why this is `cfg(test)`? I presume you're doing some sort of feature flag thing to keep this code out of the real deployment?

jayantk: fwiw I think a better place for this flag would actually be around the code the resize account code + the code that calls `add_price_to_time_machine`.

This is fine for this PR though.

jayantk (commented):
this looks very solid to me. Pretty much all of the comments are minor except the one about the `true` which i think should be `false`.

program/rust/src/time_machine_types.rs
guibescos: Yes

program/rust/src/time_machine_types.rs
guibescos: Removed

program/rust/src/time_machine_types.rs
guibescos: Yes

program/rust/src/time_machine_types.rs
guibescos: Fixed

program/rust/src/time_machine_types.rs
guibescos: Yes

program/rust/src/time_machine_types.rs
guibescos: Yes

program/rust/src/time_machine_types.rs
guibescos: Yes

program/rust/src/tests/test_sma.rs
guibescos: Yes

program/rust/src/time_machine_types.rs
guibescos: I think you're right

program/rust/src/time_machine_types.rs
guibescos: Added some comments. We can definitely gate those two instructions, but there's a test in the python for resizing that would need to get scrapped (the python tests use the production version of the rust binary).

guibescos:
`current_epoch_is_valid` gets initialized to false now.

jayantk (approved):
nice!

PR #293, opened 2022-09-13
Improve errors

guibescos:
Replace `InvalidArgument` by `InvalidNumberOfAccounts` and `InvalidAccountHeader`.

### DIFF
--- a/program/rust/src/error.rs
+++ b/program/rust/src/error.rs
@@ -40,6 +40,10 @@ pub enum OracleError {
     InvalidUpgradeAuthority        = 614,
     #[error("FailedPdaVerification")]
     InvalidPda                     = 615,
+    #[error("InvalidAccountHeader")]
+    InvalidAccountHeader           = 616,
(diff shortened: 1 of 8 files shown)


program/rust/src/error.rs
```diff
@@ -40,6 +40,10 @@ pub enum OracleError {
     InvalidUpgradeAuthority        = 614,
     #[error("FailedPdaVerification")]
     InvalidPda                     = 615,
+    #[error("InvalidAccountHeader")]
+    InvalidAccountHeader           = 616,
+    #[error("InvalidNumberOfAccounts")]
```
tompntn, line 45: nit: the `error` attribute is supposed to be a human-readable explanation of the error - for example `invalid number of accounts passed`. This representation can then be used in logs and user-facing error messages, without the need to map the errors to strings.

tompntn (approved):
Nice!

jayantk (approved):

PR #296, opened 2022-09-22
Extra check from audit

guibescos:
Check on the funding account from the internal audit.
I don't think there's ever a situation where funder is not owned by the system program. Even `executor_key` from the executor will be owned by system program.

### DIFF
--- a/program/rust/src/utils.rs
+++ b/program/rust/src/utils.rs
@@ -29,6 +29,7 @@ use solana_program::system_instruction::{
     assign,
     transfer,
 };
+use solana_program::system_program;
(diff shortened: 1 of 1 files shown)


njkumr (commented):
checking for if the the account owner is a system program could potentially lead to the instruction also erroring out if the funding account is a multi-sig. Is this something we are okay with?

guibescos:
I don't think that's true. They'll be PDAs of the multisig, but owned by the system program.

guibescos:
Closing because it seems to restrictive

PR #297, opened 2022-09-23
Guibescos/transition to permissions

guibescos:
Update all instructions that require account signatures to having the `PermissionAccount` checks.
It is backwards compatible and this PR doesn't try to define the permissions.

### DIFF
--- a/program/rust/src/tests/test_permission_migration.rs
+++ b/program/rust/src/tests/test_permission_migration.rs
@@ -3,14 +3,9 @@ use crate::c_oracle_header::{
     PermissionAccount,
     PriceAccount,
     ProductAccount,
-    PC_ACCTYPE_MAPPING,
-    PC_MAGIC,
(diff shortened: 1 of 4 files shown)


program/rust/src/tests/test_permission_migration.rs
```diff
@@ -0,0 +1,268 @@
...
+        &[
+            funding_account.clone(),
+            mapping_account.clone(),
+            permissions_account.clone(),
+        ],
+        bytes_of::<CommandHeader>(&InitMapping.into()),
+    )
+    .unwrap();
+
+    assert_eq!(
```
jayantk, line 102: I'm confused about the tests below. I thought that `funding_account` is permissioned to do these actions, and the `attacker_account` isn't. It looks like every test below is a permission violation

Maybe leave some code comments explaining what the tests are supposed to be testing.

PR #299, opened 2022-10-25
Apply Clippy & Rustfmt to project.

Reisen:
This PR applies 3 changes:

- Clippy applied to tests.
- Updates all `.rs` files to match clippy changes.
- Updates `rustfmt.toml` with settings that improve readibility.

Note: this relies on `cargo +nightly rustfmt` which is common when relying on options `rustfmt` won't include in `stable`, several Rust projects follow this practice. The CI pre-commits are already checking using `rustfmt` nightly as well. Editors may need configuring but should be able to do this by default.


guibescos (dismissed):
This is ready

guibescos (approved):

PR #300, opened 2022-10-25
Re-organize Modules

Reisen:
This PR re-organizes many of the modules in the current Rust rewrite of the oracle contract, in particular:

## 1: Creates `accounts.rs` for `PythAccount`

This moves `PythAccount` into its own module, and moves methods from `utils` that are parameterized by unconstrained `T`, this allows the methods to be correctly bound to their `impls`, all call-sites change in this way:

```diff
- initialize_pyth_account_checked::<MappingAccount>(&account_info)?
+ MappingAccount::initialize(&account_info)?
```

All account specific helpers from `utils.rs` are then moved to `accounts.rs` to move away from the `utils.rs` anti-pattern. Each individual account exists under `accounts/` which also means associated data can be colocated with the account, such as the `PERMISSIONS` seed which now lives in `accounts/permission.rs`.

## 2: Moves handlers under `processor/`

This removes the `rust_oracle` module and moves handlers under the `processor` module, each handler has its own module which makes it easier to colocate functions that are only related to specific handlers. For example the `c_upd_aggregate` function now lives in `processor/upd_price.rs`

These two changes nearly empty `c_oracle_header.rs` which was an ad-hoc place for account definitions and utilities. The rest of `utils.rs` can also be refactored, but as a separate PR.


guibescos (approved):
Let's do it

PR #302, opened 2022-11-02
Bump Solana Version to fix CI issue

jayantk:
This PR changes:
- Solana rust dependency versions to 1.13.3 which is the latest mainnet release.
- Uses Solana cli tools of v1.14.7 that have sbf instead of bpf. It seems that since bpf is outdated, for whatever reason, crates.io index cannot detect a suitable dependency for time that matches the flags/architecture. This change solves that problem.


ali-bahjati (approved):
Awesome!

PR #306, opened 2022-11-10
Make requested CU amount and price in upd_price transactions configurable in pythd

tompntn:
This allows publishers to configure the amount and cost of CU units used in `upd_price` transactions. By default we specify the amount, but not the cost.

Note that one code path, that used by the `pyth upd_price_val` cli command, used only in tests, uses a slightly different method of sending the transaction than the websocket `update_price` command. To ensure this still works, we preserve the signatures of the existing `pc::rpc::upd_price::build` and `pc::rpc::upd_price::request` methods, instead adding additional versions of these that allow specifying the additional parameters.


jayantk (approved):
seems great to me. one request is to document the special semantics of 0 in an appropriate place. Aside from that, lgtm.

jayantk (approved):

PR #317, opened 2022-11-28
Make PDA init simpler

guibescos:
97k -> 92k
More details on ptal.


jayantk (approved):
i guess this is fine -- even if someone exploits this, we can always update the code to handle it.

i suggest writing down a list of PRs to revert once the code size issue gets fixed. It's a little unfortunate to lose our robustness to these kinds of issues.

PR #318, opened 2022-11-28
Fix solana dir

guibescos:
The goal of this PR is fixing an annoying quirk of the repo that meant you had to the `pyth_client` repo and the `solana` repo side to side in the same folder. The trick is we can find the solana local installation by doing `which cargo-build-bpf`. This is inspired by https://github.com/guibescos/example-helloworld .

Resolving the solana installation ended up a little bit dirty in `build.rs` because bindgen needs to find the solana C headers. This will get better when we get rid of the C headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,6 @@ root@pyth-dev# usermod -u 1002 -g 1004 -s /bin/bash pyth
 
 Finally, in docker extension inside VS Code click right and choose "Attach VS Code". If you're using a remote host in VS Code make sure to let this connection be open.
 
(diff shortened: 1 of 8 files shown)


jayantk (approved):
nice. huge quality of life improvement

jayantk:
oh can you take a look at the docs and update anything about the side-by-side solana dir?

guibescos:
> 

Updated docs and all the other places where the path of solana was needed

jayantk (approved):

PR #319, opened 2022-11-28
Use custom bpf-tools

guibescos:
With the current solana-labs bpf-tools (the toolchain that builds bpf programs), the flags `-Z build-std=std,panic_abort` don't work. These flags compile the standard rust library from scratch and make the program smaller. 

There is evidence that the solana team plans to release a version that support this : https://github.com/solana-labs/solana/blob/9f370475d40dbe68f1a51d34fc3df07e5327d6c3/sdk/cargo-build-sbf/src/main.rs#L625

This PR is meant to make the repo use a fork of bpf-tools instead of the solana-labs release. The only difference between the my fork and solana-labs is that my fork uses https://github.com/solana-labs/compiler-builtins/tree/sbf-tools-v1.30 instead of  https://github.com/solana-labs/compiler-builtins/tree/bpf-tools-v1.27 (a package bump basically) . 

TLDR : I think the solana team is gonna make a release eventually enabling what I want to do (the changes have made it to the master branch of solana but not the releases yet) but I don't really want to wait.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -62,7 +62,7 @@ RUN pip3 install solana
 # Build off-chain binaries.
 RUN cd pyth-client && ./scripts/build.sh
 
-
+RUN ./pyth-client/scripts/get-bpf-tools.sh
 # Copy solana sdk and apply patch.
 RUN mkdir solana
 RUN cp -a /usr/bin/sdk solana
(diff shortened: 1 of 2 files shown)


ptaffet (approved):
We should consider moving these repos out of guibescos if Solana takes a long time to merge these, but for now this looks good to me

PR #321, opened 2022-11-29
Add compilation flags and check size in CI

guibescos:
The goal of this PR : 
- Add the flags `-Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort` to the compilation of the oracle. This flags compile the rust std library instead of using a rebuilt binary and disable panic tracebacks.
- Add a script that checks the size of the program for CI against the hardcoded size of the onchain account.

Once this PR gets merged, it'll fit onchain!

### DIFF
--- a/scripts/check-size.sh
+++ b/scripts/check-size.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+#
+# While Solana doesn't support resizing programs, the oracle binary needs to be smaller than 81760 bytes
+ORACLE_SIZE=$(wc -c ./target/deploy/pyth_oracle.so | awk '{print $1}')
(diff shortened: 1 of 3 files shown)


jayantk (approved):

scripts/check-size.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#
+# While Solana doesn't support resizing programs, the oracle binary needs to be smaller than 81760 bytes
+ORACLE_SIZE=$(wc -c ./target/deploy/pyth_oracle.so | awk '{print $1}')
```
Reisen, line 4: Can also `$(stat -c%s ./target/deploy/pyth_oracle.so)` alternatively.

Reisen (approved):

PR #325, opened 2022-12-12
Add accounts and types to idl.json

MHHukiewitz:
As proposed by @ricardocr987 on https://github.com/aleph-im/solana-indexer-library/blob/main/packages/indexer-generator/idl/pyth.json


MHHukiewitz:
It seems to fail on some code in the branch I did not touch. Should I instead create a PR with this IDL for the main branch?

jayantk:
@guibescos will take a look at the IDL when he gets a moment.

Re: the pre-commit error, don't worry about it. This happens to us occasionally because we're using rust nightly to run the check, and they change what syntax is accepted rather frequently. I'll send a PR to fix it in main and that should resolve your issue here.

guibescos:
I'm going to close this one.

1. We've moved the idl to https://github.com/pyth-network/pyth-client-js/blob/main/src/anchor/idl.json .
2. The new idl has the instructions but not the accounts would be awesome to merge in the accounts that came in this PR, however I'd like to have some unit tests, like the ones I wrote in  https://github.com/pyth-network/pyth-client-js/blob/main/src/__tests__/Anchor.test.ts

PR #327, opened 2022-12-21
Default value of 20 for min pub

guibescos:
A lot of price feeds have been created in the past with min_pub equal to 0, leading to protocols thinking the price is live when it's not.

The goal is changing the default behavior so it's less error prone. By default price accounts will have min_pub equal to 20.

### DIFF
--- a/program/rust/src/processor/add_price.rs
+++ b/program/rust/src/processor/add_price.rs
@@ -76,6 +76,7 @@ pub fn add_price(
     price_data.price_type = cmd_args.price_type;
     price_data.product_account = *product_account.key;
     price_data.next_price_account = product_data.first_price_account;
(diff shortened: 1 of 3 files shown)


program/rust/src/processor/add_price.rs
```diff
@@ -76,6 +76,7 @@ pub fn add_price(
     price_data.price_type = cmd_args.price_type;
     price_data.product_account = *product_account.key;
     price_data.next_price_account = product_data.first_price_account;
+    price_data.min_pub_ = 20;
```
jayantk, line 79: suggest making 20 a constant variable somewhere and then referencing that variable here

jayantk (dismissed):

jayantk (approved):

PR #332, opened 2023-01-24
[pyth_oracle] Allow add_product to create product metadata

guibescos:
Right now creating a product and adding metadata requires `add_product` and `upd_product`. This PR aims to allow initializing the metadata when you call `add_product`. 
It is backwards compatible. If you call `add_product` the old way, the program will see that as empty metadata so the behavior will be the same as previously.

This PR is meant to enable some deterministic keys for product accounts in the future.

### DIFF
--- a/program/rust/src/processor/add_product.rs
+++ b/program/rust/src/processor/add_product.rs
@@ -1,6 +1,7 @@
 use {
     crate::{
         accounts::{
+            update_product_metadata,
             MappingAccount,
             ProductAccount,
             PythAccount,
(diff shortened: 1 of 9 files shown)


program/rust/src/processor/add_product.rs
```diff
@@ -83,5 +84,7 @@ pub fn add_product(
     )? + mapping_data.number_of_products
         * try_convert::<_, u32>(size_of::<Pubkey>())?;
 
+    update_product_metadata(instruction_data, new_product_account, hdr.version)?;
```
jayantk, line 87: Document the instruction data format in the doc comment for this function.

jayantk (approved):
nice

jayantk (approved):

PR #343, opened 2023-04-06
[pyth_oracle] Improve build.rs robustness and expose types.

Reisen:
This PR adds a feature flag, `library` to the `pyth-oracle` program. When toggled on, the library exposes its account types for use as a library instead of a standalone program. When the flag is enabled, the `build.rs` script will also switch to execute the C compilation as well. The artefacts are output within the Rust target sandbox rather than into the relative source directory (`./program/c/target`).

The benefit of this:

- The build is now properly sandboxed within the Rust build directory.
- The C is built automatically with Rust.
- The library can now be used by other programs, and the C builds successfully.

Rather than fully switch to this, the feature flag is added so any existing reliance on the current build method continues to work. The primary reason I wanted this is because I am relying on `pyth-sdk-rs` and I found it very hard to tell exactly what is different there versus the accounts defined within the contract. With this feature flag it is possible for me to enforce extra constraints in `pyth-sdk-rs` that prevent the libraries going out of sync and highlights the differences in a testable way.

See [the associated PR](https://github.com/pyth-network/pyth-sdk-rs/pull/104) for context.


program/rust/build.rs
```diff
@@ -7,7 +7,33 @@ use {
 };
 
 fn main() {
-    println!("cargo:rustc-link-search=./program/c/target");
+    // Cargo exposes ENV variables for feature flags that can be used to determine whether to do a
+    // self-contained build that compiles the C components automatically.
+    if std::env::var("CARGO_FEATURE_LIBRARY").is_ok() {
```
jayantk, line 12: is there any reason not to do this for both build paths? this seems like a great simplification of the current build process.

jayantk (dismissed):
very cool.

guibescos (dismissed):
LGTM, please fix ci

program/rust/build.rs
Reisen: No reason, this is the correct way to do it. But I wasn't sure how many things would break and this was a safe small PR to get a review on. I can make it the default and check anywhere that might break / has been relying on relative paths. I just didn't plan to do it in this PR.

guibescos: Let's get this merged and I can make it the default

guibescos (approved):

PR #344, opened 2023-04-07
Allow negative prices in aggregation

jayantk:
We want negative prices for yields. The only blocker for negative numbers is an overly-aggressive `if` statement in the aggregation logic designed to prevent overflows. I updated this if statement to the minimum necessary conditions to prevent overflow.

Note that the aggregation algorithm (`price_model`) is already tested with negative numbers (see test_price_model.c) so this should be fine. I also added a test to make sure negative numbers get aggregated properly, and changed some of the existing test cases that include negative numbers.

There was also a temporary check in the ema calculation that prevented negative numbers. I removed that check and added reasonably extensive tests for negative emas to validate the fix.

### DIFF
--- a/program/c/src/oracle/upd_aggregate.h
+++ b/program/c/src/oracle/upd_aggregate.h
@@ -180,12 +180,14 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
       int64_t price     = iptr->agg_.price_;
       int64_t conf      = ( int64_t )( iptr->agg_.conf_ );
       if ( iptr->agg_.status_ == PC_STATUS_TRADING &&
-           (int64_t)0 < conf && (conf + INT64_MIN) < price && price <= (INT64_MAX-conf) && // No overflow for INT64_MAX-conf as conf>0
+           // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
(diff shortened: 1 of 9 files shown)


program/c/src/oracle/upd_aggregate.h
```diff
@@ -102,19 +102,12 @@ static void upd_ema(
     // compute numer/denom and new value from decay factor
     pd_load( numer, ptr->numer_ );
     pd_load( denom, ptr->denom_ );
-    if ( numer->v_ < 0 || denom->v_ < 0 ) {
```
jayantk, line 105: this code was a temporary thing when we changed up the ema early on. doesn't need to exist now, and also prevents negative ema values from working properly.

program/rust/test_data/aggregation/2.result
```diff
@@ -1 +1 @@
-{"exponent":-3,"price":0,"conf":0,"status":"unknown"}
```
jayantk, line 1: all of these changed tests have input prices that are either 0 or negative

program/rust/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 // Allow non upper case globals from C
 #![allow(non_upper_case_globals)]
 
+extern crate core;
```
guibescos, line 5: I don't get this

guibescos (dismissed):

PR #346, opened 2023-04-07
Move data-driven aggregation logic tests to rust

jayantk:
I'm so tired of how broken the build / test setup in this repo is. I'm going to fix it.

This is phase one: move the data-driven aggregation logic tests to run in Rust instead of via a Python file that runs a C program (ugh).


guibescos (approved):
This is awesome, thanks so much Jayant.

PR #347, opened 2023-04-07
Move TWAP data-driven tests to rust

jayantk:
Move the TWAP data-driven test to rust also. It's actually a test of the EMA so fix the name while i'm at it.


program/c/src/oracle/upd_aggregate.h
```diff
@@ -131,8 +131,10 @@ static void upd_ema(
 }
 
 static inline void upd_twap(
-    pc_price_t *ptr, int64_t nslots, pc_qset_t *qs )
+    pc_price_t *ptr, int64_t nslots )
```
jayantk, line 134: trivial refactor: the qset argument contains a table of constants that only gets used in upd_twap, but was being passed in from upd_aggregate. Moving the table creation into upd_twap made it easier to expose the functions ^ for testing.

program/rust/src/tests/test_ema.rs
```diff
@@ -0,0 +1,109 @@
...
+        price_account.agg_.conf_ = input.conf;
+
+        upd_twap(&mut price_account, input.nslots);
+
+        assert_eq!(expected_output.twap, price_account.twap_.val_);
+        assert_eq!(expected_output.twac, price_account.twac_.val_);
+    }
+}
+
+// TODO: put these functions somewhere more accessible
```
jayantk, line 73: will do this separately -- if i move them, i would like to replace all existing usages of the unsafe ones, of which there are quite a few

program/rust/src/processor/upd_price.rs
```diff
@@ -36,12 +36,15 @@ use {
 #[link(name = "cpyth-bpf")]
 extern "C" {
     pub fn c_upd_aggregate(_input: *mut u8, clock_slot: u64, clock_timestamp: i64) -> bool;
+    pub fn c_upd_twap(_input: *mut u8, nslots: i64);
```
guibescos, line 39: Need `#[allow(unused)]` here or for you to move this to the tests.

pyth/tests/test_twap.py
```diff
@@ -1,21 +0,0 @@
-from tests.conftest import BaseTest
```
guibescos, line 1: I think you can cleanup BaseTest from `conftest.py` for this PR

program/c/src/oracle/upd_aggregate.h
```diff
@@ -18,11 +18,10 @@ typedef struct pc_qset
   pd_t      weight_[PC_COMP_SIZE];
   int64_t   decay_[1+PC_MAX_SEND_LATENCY];
   int64_t   fact_[PC_FACTOR_SIZE];
-  int32_t   expo_;
 } pc_qset_t;
 
 // initialize quote-set temporary data in heap area
-static pc_qset_t *qset_new( int expo )
+static pc_qset_t *qset_new()
```
jayantk, line 24: This change here fixes the nondeterministic test. Turns out that pc_qset_t is shared across threads and constant *except* it stores a copy of the exponent, making it unsafe to invoke this method in parallel (which`cargo test` does by default). There's no reason it needs to store the exponent -- it's just bad software design.

fix is easy: just make it constant so it's threadsafe.

program/rust/src/tests/test_ema.rs
```diff
@@ -0,0 +1,109 @@
...
+    for (input, expected_output) in inputs.iter().zip(expected_outputs.iter()) {
+        price_account.exponent = input.expo;
+        price_account.agg_.price_ = input.price;
+        price_account.agg_.conf_ = input.conf;
+
+        upd_twap(&mut price_account, input.nslots);
+
+        assert_eq!(expected_output.twap, price_account.twap_.val_);
+        assert_eq!(expected_output.twac, price_account.twac_.val_);
+    }
```
guibescos, line 70: Should you `assert_eq!(price_account.exponent, input.expo;);` and same for price and conf as a sanity check?

PR #348, opened 2023-04-10
Improve build process

jayantk:
Update build.rs so that (1) it always builds and links the C code, and (2) it knows about its dependencies so it builds everything needed / rebuilds on change. After this PR, `cargo build` and `cargo test` should "just work" for the program.

I haven't updated the documentation yet -- I want to split things out so there's a clear pythd build vs. program build, but it's not quite there yet.


.dockerignore
```diff
@@ -1 +1,2 @@
-build
\ No newline at end of file
+build
+target
```
jayantk, line 2: drive-by: don't copy unnecessary build output into the docker image. this makes things slightly better when building docker locally.

guibescos (approved):

PR #349, opened 2023-04-13
Run oracle program C tests from Rust

jayantk:
Final step in fixing this repo. You can now simply run `cargo test` and it will compile the code (including the BPF binary) and run all of the oracle program tests. I also updated the README instructions, so please check them out and see if they work for you. 

There is technically still one oracle program test in the `pctest` folder, but that test is not very good so I don't think it makes sense to move over. It only affects the twap calculation, which we already have decent test coverage of. I may add some more TWAP tests in the future so that we can update to the new version of the calculation.


program/c/src/oracle/model/test_price_model.c
```diff
@@ -14,10 +14,7 @@ qcmp( void const * _p,
   return 0;
 }
 
-int
-main( int     argc,
-      char ** argv ) {
-  (void)argc; (void)argv;
+int test_price_model() {
```
jayantk, line 17: the changes in the C test files are:
1. export a function name so we can call them from rust (instead of main)
2. delete the noisy printf statements
3. I reduced the iteration counts on a couple of the randomized tests, because they took several minutes to run.

program/c/src/oracle/util/test_prng_battery.c
```diff
@@ -1,91 +0,0 @@
-#include <stdio.h>
```
jayantk, line 1: this file is like a super exhaustive test of the prng (which itself is only used for tests). It's not actually run as part of the existing c test suite, so just get rid of it.

docker/Dockerfile
```diff
@@ -63,8 +63,6 @@ RUN ./pyth-client/scripts/patch-solana.sh
 # Build and test the oracle program.
 RUN cd pyth-client && ./scripts/build-bpf.sh .
 RUN cd pyth-client && ./scripts/check-size.sh
-# Run aggregation logic tests
-RUN cd pyth-client && ./scripts/run-aggregation-tests.sh
```
jayantk, line 67: Note that build-bpf runs cargo test-bpf, which now runs all of these tests.

program/c/src/oracle/util/test_avg.c
```diff
@@ -76,8 +73,8 @@ main( int     argc,
 # define N 512
 
   ctr = 0;
-  for( int i=0; i<10000000; i++ ) {
```
guibescos, line 79: There's another variable called `i` inside the loop, can you fix that for clarity. For example by renaming this `i` to `iter`

guibescos (approved):
Please check out the minor comments

PR #351, opened 2023-04-21
Fix unresolved_symbol error on Mac

guibescos:
I went to the dark corner of the internet to fix this : https://macports-users.macosforge.narkive.com/Wa9Zfepg/why-ar-doesn-t-call-ranlib-on-mac
Tried one of the tips and it seems to work.


jayantk (approved):
nice

PR #354, opened 2023-04-25
Message Buffer CPI

jayantk:
We've tested this in pyth-agent with the message buffer program and it seems to work. This PR deletes the old sma tests but i guess that's fine since we're replacing it anyway.


program/c/makefile
```diff
@@ -16,7 +16,7 @@ endif
 # The all target is defined by the solana makefile included above and generates the needed .o file.
 .PHONY: cpyth-bpf
 cpyth-bpf: all
-	bash -c "ar rc $(OUT_DIR)/libcpyth-bpf.a $(OUT_DIR)/oracle/*.o"
+	bash -c "ar rcs $(OUT_DIR)/libcpyth-bpf.a $(OUT_DIR)/oracle/*.o"
```
jayantk, line 19: turns out this `s` is important for making the binary size small. (also it didn't fix the ELF loading error when i tried it 🤷 )

.pre-commit-config.yaml
```diff
@@ -9,13 +9,13 @@ repos:
   - id: check-added-large-files
 - repo: local
   hooks:
-    - id: cargo-fmt-nightly
-      name: Cargo Fmt Nightly
+    - id: cargo-fmt
```
jayantk, line 12: drive-by to fix the CI failure. luckily this doesn't cause any formatting changes in other files.

program/rust/src/processor/upd_price.rs
```diff
@@ -126,31 +163,79 @@ pub fn upd_price(
...
+    let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+    // We want to send a message every time the aggregate price updates. However, during the migration,
+    // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
+    // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
+    // will send the message.
+    if aggregate_updated {
+        price_data.message_sent_ = 0;
+    }
+
+    if let Some(accumulator_accounts) = maybe_accumulator_accounts {
```
jayantk, line 175: this logic got relatively tricky and it could definitely use a unit test. Unfortunately it's quite difficult to add such a test without the message buffer program. I'll try to put one in the integration test repo.

scripts/check-size.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 #
+
 # While Solana doesn't support resizing programs, the oracle binary needs to be smaller than 81760 bytes
+# (The available space for the oracle program on pythnet is 88429 and mainnet is 81760)
```
jayantk, line 5: i've bumped this up to the pythnet size so we can get this out for the audit. we can circle back and figure out how to split out mainnet/pythnet into separate builds.

(I think there's actually some code we can delete as well around the price account keys being signers and such, and that could help)

program/rust/src/accounts/price.rs
```diff
@@ -91,3 +94,81 @@ impl PythAccount for PriceAccount {
...
+    pub price:        i64,
+    pub conf:         u64,
+    pub exponent:     i32,
+    pub publish_time: i64,
+    pub ema_price:    i64,
+    pub ema_conf:     u64,
+}
+
+impl PriceFeedMessage {
+    pub const MESSAGE_SIZE: usize = 1 + 32 + 8 + 8 + 4 + 8 + 8 + 8;
```
guibescos, line 114: Can we add an assert for this

program/rust/src/accounts/price.rs
```diff
@@ -91,3 +94,81 @@ impl PythAccount for PriceAccount {
...
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PriceFeedMessage {
+    pub id:           [u8; 32],
+    pub price:        i64,
+    pub conf:         u64,
+    pub exponent:     i32,
+    pub publish_time: i64,
+    pub ema_price:    i64,
+    pub ema_conf:     u64,
```
guibescos, line 110: We also need to send` prev_publish_time`.

program/rust/src/processor/upd_price.rs
```diff
@@ -126,14 +163,79 @@ pub fn upd_price(
         }
     }
 
+    let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+    // We want to send a message every time the aggregate price updates. However, during the migration,
+    // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
+    // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
+    // will send the message.
```
guibescos, line 170: How are you planning to send prev_timestamp_? Since it is potentially not available anymore when the next publisher comes.

guibescos (commented):
This is probably for a future PR but worth to think about now.
There's some trickiness around including `prev_publish_timestamp` in the messages. 
Since the publisher that triggers the message send is not the same that triggered the aggregate, `price_data.prev_timestamp_` might not be available anymore when we want to send the message.

jayantk:
I added prev_timestamp to the message. I think it's actually fine, we just need to get to a place where all the publishers are sending the accumulator accounts, and then we can make the guarantee we need. See the inline comment about how this works.

program/rust/src/accounts/price.rs
```diff
@@ -98,20 +98,36 @@ impl PythAccount for PriceAccount {
...
-    pub ema_conf:     u64,
+    pub id:                [u8; 32],
+    pub price:             i64,
+    pub conf:              u64,
+    pub exponent:          i32,
+    /// The timestamp of this price update
+    pub publish_time:      i64,
+    /// The timestamp of the previous price update. This field is intended to allow users to
+    /// identify the single unique price update immediately after any moment in time:
+    /// for any time t, the unique update is the one such that prev_publish_time <= t < publish_time.
```
guibescos, line 116: `prev_publish_time < t <= publish_time` since price and conf will then be the aggregate at `publish_time` ?

program/rust/src/processor/upd_price.rs
```diff
@@ -126,14 +163,79 @@ pub fn upd_price(
...
+    }
+
+    if let Some(accumulator_accounts) = maybe_accumulator_accounts {
+        if price_data.message_sent_ == 0 {
+            // Check that the oracle PDA is correctly configured for the program we are calling.
+            let oracle_auth_seeds: &[&[u8]] = &[
+                UPD_PRICE_WRITE_SEED.as_bytes(),
+                &accumulator_accounts.program_id.key.to_bytes(),
+            ];
+            let (expected_oracle_auth_pda, bump) =
```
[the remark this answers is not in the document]
guibescos: I like conciseness

program/rust/src/accounts/price.rs
```diff
@@ -91,3 +94,101 @@ impl PythAccount for PriceAccount {
...
+    /// (which we can do once publishers have migrated over).
+    pub prev_publish_time: i64,
+    pub ema_price:         i64,
+    pub ema_conf:          u64,
+}
+
+impl PriceFeedMessage {
+    // The size of the serialized message. Note that this is not the same as the size of the struct
+    // (because of the discriminator & struct padding/alignment).
+    pub const MESSAGE_SIZE: usize = 1 + 32 + 8 + 8 + 4 + 8 + 8 + 8 + 8;
```
guibescos, line 130: I still don't really like this MESSAGE_SIZE variable.
You can use size_of() statically : 
`pub fn as_bytes(&self) -> [u8; size_of::<u8>() + size_of::<PriceFeedMessage>()]`

alternatively, add a test in test_sizes.rs

jayantk: well it's not supposed to be the size of the struct (for the reasons noted in the comment). the generative test i added does check that the message exactly fills the array (i.e., no extra bytes left after reading the fields), so this size is tested already.

guibescos (approved):

PR #357, opened 2023-04-26
chore(scripts): add script to launch docker env for building

swimricky:


jayantk (approved):
nice! do you mind also updating the readme to point people to this script

scripts/start_docker_build_env.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+command -v shellcheck >/dev/null && shellcheck "$0"
+
+dir=$(pwd)
+parentdir="$(dirname "$dir")"
+export PYTH_REPO=$parentdir
+
+# Use the latest pythd image from https://hub.docker.com/r/pythfoundation/pyth-client/tags?name=pythd
```
guibescos, line 9: Can you update this to use the `oracle` instead of the `pythd` tag

scripts/start_docker_build_env.sh
```diff
@@ -0,0 +1,19 @@
...
+LATEST_VERSION="v2.17.3"
+export IMAGE=pythfoundation/pyth-client:pythd-${LATEST_VERSION}
+
+docker run -it \
+        --volume "${HOME}:/home/pyth/home" \
```
guibescos, line 14: It feels a bit excessive to mount the whole home dir. Is this intentational

scripts/start_docker_build_env.sh
```diff
@@ -0,0 +1,19 @@
...
+        --volume "${HOME}/.config:/home/pyth/.config" \
```
guibescos, line 15: This is redundant since `${HOME}/.config` is inside `${HOME}`

scripts/start_docker_build_env.sh
swimricky: Thanks, i removed this. It was copied from the "Development Setup Using VSCode" section in the README. i don't think it's necessary if the only purpose is to be able to build the oracle program from the mounted directories.

scripts/start_docker_build_env.sh
swimricky: also removed this

guibescos (approved):

PR #359, opened 2023-04-26
Revisit twap2

guibescos:


program/c/src/oracle/oracle.h
```diff
@@ -199,8 +199,8 @@ typedef struct pc_price
 static_assert( sizeof( pc_price_t ) == 3312, "" );
 
 // This constant needs to be an upper bound of the price account size, it is used within pythd for ztsd.
-// It is set tighly to the current price account + 96 component prices + 42 bytes for cumulative sums
-const uint64_t PRICE_ACCOUNT_SIZE = 3312 + 96 * sizeof( pc_price_comp_t) + 42;
```
guibescos, line 203: Typo in previous PR

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
     pub comp_:              [PriceComponent; PC_COMP_SIZE as usize],
 }
 
+/// This is the new version of the PriceAccount.
+/// It will go live once we resize all the price accounts.
+/// It will include more publisher spots and will store cumulative sums to compute TWAPS.
+#[cfg(test)]
+#[repr(C)]
+#[derive(Copy, Clone, Pod, Zeroable)]
+pub struct PriceAccountNew {
```
jayantk, line 71: nitpick: how about PriceAccountV2

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
...
+    /// Aggregate price at prev_slot_
+    pub prev_price_:        i64,
+    /// Confidence interval at prev_slot_
+    pub prev_conf_:         u64,
+    /// Timestamp of prev_slot_
+    pub prev_timestamp_:    i64,
+    /// Last attempted aggregate results
+    pub agg_:               PriceInfo,
+    /// Publishers' price components
+    pub comp_:              [PriceComponent; PC_COMP_SIZE as usize],
```
jayantk, line 111: PC_COMP_SIZE is the same number as the old price account (?)

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
...
+    /// Cumulative sums of aggregative price and confidence used to compute arithmetic moving averages
+    pub price_cumulative:   PriceCumulative,
+}
+
+#[cfg(test)]
+impl PriceAccountNew {
+    /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
+    pub fn update_price_cumulative(&mut self) {
```
jayantk, line 119: aren't there some assumptions about trading and stuff that need to be satisfied in order to call this function?

IMO this function does so little it would be better to inline the call to update in the calling code (which also guarantees the assumptions are satisfied)

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
...
+            self.agg_.conf_,
+            self.agg_.pub_slot_.saturating_sub(self.prev_slot_), // pub_slot should always be >= prev_slot, but we protect ourselves against underflow just in case
+        );
+    }
+}
+
+// This struct can't overflow since :
+// |sum(price * slotgap)| <= sum(|price * slotgap|) <= max(|price|) * sum(slotgap) <= i64::MAX * * current_slot <= i64::MAX * u64::MAX <= i128::MAX
+// |sum(conf * slotgap)| <= sum(|conf * slotgap|) <= max(|conf|) * sum(slotgap) <= u64::MAX * current_slot <= u64::MAX * u64::MAX <= u128::MAX
+// num_gaps <= current_slot <= u64::MAX
```
jayantk, line 131: 👍 nice documentation

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
...
+#[cfg(test)]
+#[repr(C)]
+#[derive(Copy, Clone, Pod, Zeroable)]
+pub struct PriceCumulative {
+    pub price:    i128, // Cumulative sum of price * slot_gap
+    pub conf:     u128, // Cumulative sum of conf * slot_gap
+    pub num_gaps: u64,  // Cumulative number of gaps in the uptime
+    pub unused:   u64,  // Padding for alignment
```
jayantk, line 139: don't we also need the slot number to calculate the TWAP? is that just `last_slot_` in the price account? 

one way to address this is to include the procedure for calculating the TWAP in the comments somewhere

program/rust/src/tests/test_twap.rs
```diff
@@ -0,0 +1,143 @@
...
+            price:    1,
+            conf:     2,
+            slot_gap: 4,
+        },
+        DataEvent {
+            price:    i64::MAX,
+            conf:     u64::MAX,
+            slot_gap: 1,
+        },
+    ];
```
jayantk, line 112: maybe add a test with a negative price just to be explicit. (i know the generative test covers it, but it's a little cheat-y to test the code against exactly the same implementation)

program/rust/src/processor/upd_price.rs
```diff
@@ -126,6 +131,18 @@ pub fn upd_price(
         }
     }
 
+    #[cfg(test)]
```
jayantk, line 134: this doesn't need `cfg(test)` right? like it only triggers once the account is resized, and right now they haven't been resized?

program/rust/src/accounts/price.rs
jayantk: it's a little concerning that this error isn't caught by the size tests below. something seems wrong there as well.

program/rust/src/accounts/price.rs
```diff
@@ -91,3 +182,10 @@ impl PythAccount for PriceAccount {
     /// Equal to the offset of `comp_` in `PriceAccount`, see the trait comment for more detail
     const INITIAL_SIZE: u32 = PC_PRICE_T_COMP_OFFSET as u32;
 }
+
+#[cfg(test)]
+impl PythAccount for PriceAccountNew {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRICE;
+    /// Equal to the offset of `comp_` in `PriceAccount`, see the trait comment for more detail
+    const INITIAL_SIZE: u32 = PC_PRICE_T_COMP_OFFSET as u32;
```
jayantk, line 190: is this right? I think you've allocated some bytes *after* the component array now, so the initial size needs to include the entire array + the subsequent values?

program/rust/src/tests/test_upd_price.rs
```diff
@@ -41,13 +41,13 @@ fn test_upd_price() {
     let mut funding_setup = AccountSetup::new_funding();
     let funding_account = funding_setup.as_account_info();
 
-    let mut price_setup = AccountSetup::new::<PriceAccount>(&program_id);
```
jayantk, line 44: i suggest copy/pasting this test code to also test the vanilla `PriceAccount` codepath. Both of them should work for the moment (and we can delete the copy/paste when we delete the old PriceAccount)

program/rust/src/accounts/price.rs
```diff
@@ -60,6 +62,95 @@ pub struct PriceAccount {
     pub comp_:              [PriceComponent; PC_COMP_SIZE as usize],
 }
 
+/// This is the new version of the PriceAccount.
+/// It will go live once we resize all the price accounts.
+/// It will include more publisher spots and will store cumulative sums to compute TWAPS.
```
jayantk, line 67: can you provide more context for this? tell the whole story "we are out of space in price account and in the process of reallocating the accounts...'

jayantk: also like "while both versions of these account structs are available, all operations on price accounts *must* work when passed either version. Operations may check the size of the account to determine which version of the account they have."

jayantk (commented):
awesome. I have a couple asks for this one, so would like to see this again after some updates.

program/rust/src/accounts/price.rs
guibescos: This was intentional, but I agree it's very confusing. Added the extra pubs.

program/rust/src/accounts/price.rs
guibescos: The reason it's not inlined is some rust borrower annoyance that makes it not very clean.
I added a check that status == TRADING whenever this function gets called.

.github/workflows/docker.yaml
```diff
@@ -2,6 +2,7 @@ name: Docker
 
 on:
   push:
+    branches: [ main ]
```
guibescos, line 5: Drive-by reduce CI load

program/c/src/oracle/oracle.h
```diff
@@ -200,7 +201,7 @@ static_assert( sizeof( pc_price_t ) == 3312, "" );
 
 // This constant needs to be an upper bound of the price account size, it is used within pythd for ztsd.
 // It is set tighly to the current price account + 96 component prices + 48 bytes for cumulative sums
-const uint64_t PRICE_ACCOUNT_SIZE = 3312 + 96 * sizeof( pc_price_comp_t) + 48;
+const uint64_t ZSTD_UPPER_BOUND = 3312 + 96 * sizeof( pc_price_comp_t) + 48;
```
guibescos, line 204: This get renamed

scripts/check-size.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 #
+
 # While Solana doesn't support resizing programs, the oracle binary needs to be smaller than 81760 bytes
+# (The available space for the oracle program on pythnet is 88429 and mainnet is 81760)
 ORACLE_SIZE=$(wc -c ./target/deploy/pyth_oracle.so | awk '{print $1}')
-if [ $ORACLE_SIZE -lt 81760 ]
+if [ $ORACLE_SIZE -lt 88429 ]
```
guibescos, line 7: I think we the c flags we can stay under this with the 3 upcoming changes : twap + resize + message-buffer

program/rust/src/accounts/price.rs
```diff
@@ -91,3 +193,9 @@ impl PythAccount for PriceAccount {
     /// Equal to the offset of `comp_` in `PriceAccount`, see the trait comment for more detail
     const INITIAL_SIZE: u32 = PC_PRICE_T_COMP_OFFSET as u32;
 }
+
+impl PythAccount for PriceAccountV2 {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PRICE;
+    /// Equal to the offset of `comp_` in `PriceAccount`, see the trait comment for more detail
```
jayantk, line 199: comment is wrong now

jayantk (approved):
awesome. i think there's a minor issue with the header size still, but should be an easy fix.

jayantk (dismissed):

guibescos:
This is now blocked by https://github.com/pyth-network/pyth-client/pull/361

jayantk (approved):

PR #361, opened 2023-04-27
[size] Add -Oz flag to further optimize c code.

guibescos:


jayantk (commented):
Seems good to me. I think it's worth seeing what @Reisen thinks because i don't really understand what the consequences of changing these flags are (besides smaller code)

jayantk (approved):

PR #362, opened 2023-05-01
[size-hack] Remove clear account

guibescos:
I need to make to program smaller again. This is one idea.

Everytime we create a new account, we manually zero it. 
I think this is and has always been unnecessary. Being able to create non-zero accounts belonging to a given program would break solana.

### DIFF
--- a/program/rust/src/accounts.rs
+++ b/program/rust/src/accounts.rs
@@ -20,16 +20,20 @@ use {
         account_info::AccountInfo,
         program::invoke_signed,
         program_error::ProgramError,
-        program_memory::sol_memset,
         pubkey::Pubkey,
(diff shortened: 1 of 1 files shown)


program/rust/src/accounts.rs
```diff
@@ -109,7 +113,6 @@ pub trait PythAccount: Pod {
         )?;
 
         check_valid_fresh_account(account)?;
-        clear_account(account)?;
```
jayantk, line 112: maybe worth a code comment that the solana runtime guarantees that the account content is all zero here.

(feel free to do in a later pr because i know the CI takes forever here)

jayantk (approved):

PR #364, opened 2023-05-05
Add TWAP Message format

guibescos:
There's some awkwardness around whether the messages get created from PriceAccount or PriceAccoutV2 (the twap fields only exist in PriceAccountV2) but we can work on that later. Right now accounts have to be upgraded before messages start getting sent.

Key point is the `TwapMessage` struct.

### DIFF
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -11,14 +11,22 @@ use {
             PC_MAX_SEND_LATENCY,
             PC_STATUS_TRADING,
         },
+        deserialize::load_checked,
         error::OracleError,
     },
     bytemuck::{
         Pod,
         Zeroable,
     },
-    solana_program::pubkey::Pubkey,
-    std::mem::size_of,
+    solana_program::{
+        account_info::AccountInfo,
+        program_error::ProgramError,
+        pubkey::Pubkey,
+    },
+    std::{
+        cell::RefMut,
+        mem::size_of,
+    },
 };
 
 #[repr(C)]
@@ -96,7 +104,7 @@ pub struct PriceAccountV2 {
     pub timestamp_:         i64,
     /// Minimum valid publisher quotes for a succesful aggregation
     pub min_pub_:           u8,
-    pub unused_1_:          i8,
+    pub message_sent_:      u8,
     pub unused_2_:          i16,
     pub unused_3_:          i32,
     /// Corresponding product account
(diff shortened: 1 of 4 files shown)


program/rust/src/accounts/price.rs
```diff
@@ -240,7 +241,7 @@ impl PriceFeedMessage {
     pub const MESSAGE_SIZE: usize = 1 + 32 + 8 + 8 + 4 + 8 + 8 + 8 + 8;
     pub const DISCRIMINATOR: u8 = 0;
 
-    pub fn from_price_account(key: &Pubkey, account: &PriceAccount) -> Self {
+    pub fn from_price_account(key: &Pubkey, account: &PriceAccountV2) -> Self {
```
guibescos, line 244: I switched the messages to be derived from PriceAccountV2, otherwise the code was getting too complicated.

program/rust/src/accounts/price.rs
```diff
@@ -302,3 +303,82 @@ impl PriceFeedMessage {
...
+#[repr(C)]
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct TwapMessage {
+    pub id:                [u8; 32],
+    pub cumulative_price:  i128,
+    pub cumulative_conf:   u128,
+    pub num_down_slots:    u64,
+    pub exponent:          i32,
+    pub publish_time:      i64,
+    pub prev_publish_time: i64,
```
ali-bahjati, line 317: this is to make sure we get the unique one per second?

ali-bahjati (approved):
LGTM ser. I have some questions that I asked in inline comments.

program/rust/src/accounts/price.rs
guibescos: Yes

program/rust/src/accounts/price.rs
```diff
@@ -302,3 +303,82 @@ impl PriceFeedMessage {
...
+        };
+
+        Self {
+            id: key.to_bytes(),
+            cumulative_price: account.price_cumulative.price,
+            cumulative_conf: account.price_cumulative.conf,
+            num_down_slots: account.price_cumulative.num_down_slots,
+            exponent: account.exponent,
+            publish_time,
+            prev_publish_time: account.prev_timestamp_,
```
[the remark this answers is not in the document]
ali-bahjati: ok i understood why it won't be equal in the future. i think it's good to make a comment about this assumption.

ali-bahjati (approved):

PR #367, opened 2023-05-08
Remove add_mapping

guibescos:
I think this function has an identity crisis. It is normally use to create additional mapping accounts. But I'd like to keep 1 mapping account and resize it if we need more space.

Since I'm running out of on-chain space I propose deleting it now.


ali-bahjati (approved):
Tyvm

PR #368, opened 2023-05-09
Conditional compilation

guibescos:
Add conditional compilation with the feature flag pythnet. We now check size for both pythnet and mainnet.


ali-bahjati (approved):
It makes a lot of sense. 🙌

PR #372, opened 2023-05-24
Add deserialization to price messages

ali-bahjati:
This PR adds deserialization to PriceFeedMessage and TwapMessage. It also introduces a Message enum which wraps messages. I have refactored the code such as renaming `as_bytes` to `to_bytes` because `as_bytes` usually returns a reference from a type that is wrapping bytes.

The purpose of this change is to use these structs from this crate (as library) in hermes.


jayantk (dismissed):
lgtm, though i have one minor ask to help us keep the program size down in the future.

guibescos (dismissed):

cctdaniel (approved):

PR #373, opened 2023-05-26
Make adding and deleting pubs idempotent

guibescos:


program/rust/src/tests/test_add_publisher.rs
```diff
@@ -128,6 +145,6 @@ fn test_add_publisher() {
             &[funding_account.clone(), price_account.clone(),],
             instruction_data
         ),
-        Err(ProgramError::InvalidArgument)
+        Ok(())
```
jayantk, line 148: maybe check here that the pub array is unchanged after this operation?

jayantk (commented):
I think this change is fine, though the case where you hit the 32 pub limit is weird. I get why you're doing that though, and it is technically idempotent so fine. I think my one ask is to document the semantics of both the add and delete operations so it's clear to the reader why the functions return `Ok()` in these cases where you might expect an error.

PR #375, opened 2023-06-02
Improvements to Message in the Oracle Program

Reisen:
This PR extends the `Message` type with features that are helpful to users consuming the program as a library.

- Adds strum/serde to `Message*` during `--features = library` use.
- Adds methods to extract id/publish_time from any message.
- Exports the new `MessageType` enum generated by strum.

### DIFF
--- a/program/rust/Cargo.toml
+++ b/program/rust/Cargo.toml
@@ -15,6 +15,8 @@ thiserror = "1.0"
 num-derive = "0.3"
 num-traits = "0.2"
 byteorder = "1.4.3"
+serde = { version = "1.0", features = ["derive"], optional = true }
+strum = { version = "0.21.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
 solana-program-test = "=1.13.3"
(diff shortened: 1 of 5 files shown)


program/rust/Cargo.toml
```diff
@@ -30,10 +32,9 @@ serde_json = "1.0"
 test-generator = "0.3.1"
 csv = "1.1"
 
-
 [features]
 debug = []
-library = []
+library = ["serde", "strum"]
```
guibescos, line 37: What is the reason for having two "subfeatures" serde and strum.
Instead of using `#[cfg(feature = "library")]` in the code

Reisen: Just for the flexibility, in reality most of the time someone is going to want both but this allows for this pattern:

```
pyth-oracle = { default-features = false, features = ["strum"] }
```

Etc.

guibescos (dismissed):

program/rust/Cargo.toml
ali-bahjati: well thinking more, no Message will be exported without `library` feature though, right?

Maybe we can keep serde separate but i was thinking maybe we can hide strum from users entirely :?

guibescos (dismissed):

ali-bahjati (approved):

PR #377, opened 2023-06-12
Use message structures defined in pythnet sdk

m30m:
Since the custom serialization implementations are solana specific and mainly for reducing the binary size, they are kept in this package via a custom trait.
Deserialization functions were removed since they were not used and the default serde provided by pythnet sdk should work identically. Serialization tests are now repurposed to test our custom serialization and the general deserialization lead to the same original structure.

### DIFF
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -17,18 +17,12 @@ use {
         Pod,
         Zeroable,
     },
-    byteorder::{
-        BigEndian,
-        ReadBytesExt,
+    pythnet_sdk::messages::{
+        PriceFeedMessage,
+        TwapMessage,
     },
     solana_program::pubkey::Pubkey,
-    std::{
-        io::{
-            Cursor,
-            Read,
-        },
-        mem::size_of,
-    },
+    std::mem::size_of,
 };
 
 #[repr(C)]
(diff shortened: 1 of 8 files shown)


program/rust/src/tests/test_message.rs
```diff
@@ -1,97 +1,40 @@
...
+    match reconstructed {
+        Message::PriceFeedMessage(reconstructed) => reconstructed == input,
+        _ => false,
     }
 }
 
 #[quickcheck]
 fn test_twap_message_roundtrip(input: TwapMessage) -> bool {
-    let reconstructed = TwapMessage::try_from_bytes(&input.to_bytes());
+    let reconstructed = from_slice::<BigEndian, Message>(&input.to_bytes()).unwrap();
```
jayantk, line 30: just to be clear about this: If the message struct changes (e.g., we add a field in the SDK), and then we upgrade the dependency here, this test will fail because the generator for the message will be upgraded, but `to_bytes()` will not be. (I think that is a good thing, but I want to make sure that is what will hapeen)

jayantk (dismissed):
nice!

program/rust/src/tests/test_message.rs
m30m: Correct

m30m:
Size changes:
- pythnet: from 86728 to 86776 < 88429
- mainnet: from 81056 to 81056 < 81760

program/rust/src/accounts/price.rs
```diff
@@ -382,115 +295,23 @@ impl PriceFeedMessage {
...
-    /// to the size of the compiled binary (which is already close to the size limit).
+impl PythOracleSerialize for TwapMessage {
+    /// Ideally, this structure should not be aware of the discrminator and the Message enum
+    /// should handle serializing the discriminator. However, this method is kept as is in order
+    /// to avoid any further increase in the program binary size since we are close to the limit.
     #[allow(unused_assignments)]
-    pub fn to_bytes(self) -> [u8; Self::MESSAGE_SIZE] {
-        let mut bytes = [0u8; Self::MESSAGE_SIZE];
+    fn to_bytes(self) -> Vec<u8> {
+        const MESSAGE_SIZE: usize = 1 + 32 + 16 + 16 + 8 + 4 + 8 + 8 + 8;
```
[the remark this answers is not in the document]
m30m: I discussed this with Ali, we were thinking of moving the DISCRIMINATOR serialization part to the Message enum and implement to_bytes for it as well as the next step. This would lead to `PythOracleSerialize` becoming more abstract and not having a fixed `discriminator`/`message_size`. That's why we kept it like this.

guibescos (dismissed):

ali-bahjati (approved):

ali-bahjati (approved):

PR #385, opened 2023-11-07
Prove that new publishing slots work in a full-pubset test

drozdziak1:
# Motivation
As yet another method for sanity-checking the increased publisher count on Pythnet, it is useful to check that the new slots are used for aggregation properly, just as well as the pre-existing ones. 

# Summary of changes
* `test_full_publisher_set.rs` - Add a new test that triggers aggregation on two subsets of a symbol's publishers, proving that for Pythnet, the new publishing slots are used for aggregation. This test is run both for Pythnet and Solana on purpose, to validate that assumpions about aggregate behavior are consistent on both of the networks.
* `utils.rs` - Rename `get_status_for_update` to `get_status_for_conf_price_ratio` to better indicate that it may result in rejecting the updated price in next aggregation.
* `upd_price.rs` - Point out that a failed `get_status_for_price_conf_ratio` may render the price update rejected from next aggregate
* `pyth_simulator.rs` - Modify some setup methods to take a slice of publisher keypairs, making multi-pub setup possible.

### DIFF
--- a/program/rust/src/processor/upd_price.rs
+++ b/program/rust/src/processor/upd_price.rs
@@ -193,11 +193,6 @@ pub fn upd_price(
         }
     }
 
-    solana_program::msg!(
-        "Aggregate {}updated",
-        if aggregate_updated { "" } else { "not " },
-    );
-
     // Reload price data as a struct after c_upd_aggregate() borrow is dropped
     let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
 
--- a/program/rust/src/processor/upd_price.rs
+++ b/program/rust/src/processor/upd_price.rs
@@ -193,6 +193,11 @@ pub fn upd_price(
         }
     }
 
+    solana_program::msg!(
+        "Aggregate {}updated",
+        if aggregate_updated { "" } else { "not " },
+    );
+
     // Reload price data as a struct after c_upd_aggregate() borrow is dropped
     let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
 
--- a/program/rust/src/processor/upd_price.rs
+++ b/program/rust/src/processor/upd_price.rs
@@ -24,7 +24,7 @@ use {
         utils::{
             check_valid_funding_account,
             check_valid_writable_account,
-            get_status_for_update,
+            get_status_for_conf_price_ratio,
             is_component_update,
             pyth_assert,
             try_convert,
@@ -276,8 +276,11 @@ pub fn upd_price(
 
     // Try to update the publisher's price
     if is_component_update(cmd_args)? {
+        // IMPORTANT: If the publisher does not meet the price/conf
+        // ratio condition, its price will not count for the next
+        // aggregate.
         let status: u32 =
-            get_status_for_update(cmd_args.price, cmd_args.confidence, cmd_args.status)?;
+            get_status_for_conf_price_ratio(cmd_args.price, cmd_args.confidence, cmd_args.status)?;
 
         {
             let publisher_price = &mut price_data.comp_[publisher_index].latest_;
--- a/program/rust/src/tests/pyth_simulator.rs
+++ b/program/rust/src/tests/pyth_simulator.rs
@@ -422,7 +422,7 @@ impl PythSimulator {
 
         for (key, price_account) in price_accounts {
             let cmd = UpdPriceArgs {
-                header:          OracleCommand::UpdPriceNoFailOnError.into(),
+                header:          OracleCommand::UpdPrice.into(),
                 status:          quotes[key].status,
                 unused_:         0,
                 price:           quotes[key].price,
@@ -536,15 +536,17 @@ impl PythSimulator {
     /// TODO : this fixture doesn't set the product metadata
     pub async fn setup_product_fixture(
         &mut self,
-        publisher: Pubkey,
+        publishers: &[Pubkey],
         security_authority: Pubkey,
     ) -> HashMap<String, Pubkey> {
         let result_file =
             File::open("./test_data/publish/products.json").expect("Test file not found");
 
-        self.airdrop(&publisher, 100 * LAMPORTS_PER_SOL)
-            .await
-            .unwrap();
+        for publisher in publishers {
+            self.airdrop(&publisher, 100 * LAMPORTS_PER_SOL)
+                .await
+                .unwrap();
+        }
 
         self.airdrop(&security_authority, 100 * LAMPORTS_PER_SOL)
             .await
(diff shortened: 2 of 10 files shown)


program/rust/src/processor/upd_price.rs
```diff
@@ -193,6 +193,11 @@ pub fn upd_price(
         }
     }
 
+    solana_program::msg!(
```
drozdziak1, line 196: ugh, I missed this, sorry!

program/rust/src/tests/pyth_simulator.rs
```diff
@@ -536,15 +536,17 @@ impl PythSimulator {
     /// TODO : this fixture doesn't set the product metadata
     pub async fn setup_product_fixture(
```
jayantk, line 537: nit: comment should say `publishers`

program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -0,0 +1,133 @@
...
+// * Publish two distinct price values to first/mid third
+// * Verify that the aggregate averages out to an expected value in the middle
+// * Publish PC_STATUS_UNKNOWN prices to first_third to make sure it does not affect later aggregates
+// * Repeat the two-price-value publishing step, using completely
+// different values, this time using mid_third and last_third without first_third
+// * Verify again for the new values meeting in the middle as aggregate.
+#[tokio::test]
+async fn test_full_publisher_set() -> Result<(), Box<dyn std::error::Error>> {
+    let mut sim = PythSimulator::new().await;
+    let pub_keypairs: Vec<_> = (0..PC_NUM_COMP).map(|_idx| Keypair::new()).collect();
```
jayantk, line 36: you sure this constant has the right value on each network? (i think you may have added a separate test toggled by the feature flag that it does have the expected value, but i forget)

jayantk (approved):
looks solid to me

program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -0,0 +1,133 @@
...
+    let n_pubs = pub_keypairs.len();
+
+    // Divide publishers into three ~even parts +/- 1 pub. The '+ 1'
+    // ensures that for Solana oracle, any two of the parts form a
+    // subset that's larger than default min_pub_ value (currently
+    // 20).
+    let (first_third, other_two_thirds) = pub_keypairs.split_at(n_pubs / 3 + 1);
+    let (mid_third, last_third) = other_two_thirds.split_at(n_pubs / 3 + 1);
+
+    for (first_kp, mid_kp) in first_third.iter().zip(mid_third.iter()) {
```
guibescos, line 56: I'd feel better if there was test where all 64 publishers publish at the same time

program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -0,0 +1,133 @@
...
+        let mid_quote = Quote {
+            price:      120,
+            confidence: 30,
+            status:     PC_STATUS_TRADING,
+        };
+
+        sim.upd_price(mid_kp, price, mid_quote).await?;
+
+        slot += 1;
+        sim.warp_to_slot(slot).await?;
```
guibescos, line 74: I don't think you need to warp slot here, the slots already move forward while you're doing the slow task of updating all prices

program/rust/src/tests/test_full_publisher_set.rs
guibescos: Here it's only 2/3 at a time

program/rust/src/tests/pyth_simulator.rs
drozdziak1: I believe this comment is correct. there's stuff like symbol name in the test case assets, but we only make it available for lookup to distinguish symbols human-readably, without filling in the relevant field in the metadata k/v/ store.

program/rust/src/tests/test_full_publisher_set.rs
drozdziak1: You are right, I am no longer sure why I went for the 2/3 scenario. I made it just all publishers, with two price values divided  50/50 between the halves.  The premise being - if not all publishers are in use, the aggregation should prefer the majority and result in an aggregate closer to it, which should break the assertion.

program/rust/src/tests/test_full_publisher_set.rs
drozdziak1: There's a couple potential problems to discuss:
* The size is not computed right - we already explicitly verify in `test_sizes.rs` that Pythnet has 64 and Solana has 32 publishers. 
* In the horrifying case that there's still a mismatch between what the C and Rust code is actually using - it should become evident as part of this test:
** If C code is using 64 and Rust is giving it 32 pubs - we would likely see an invalid memory access crash the program.
** If C code is using 32 and Rust is giving it 64 pubs - there should be an assertion failure due to aggregation working only on the first half of the publisher set, and computing a trivial aggregate price of `100` with confidence `30`, instead of the `110` with conf `20` that's supposed to come out of the two values "meeting in the middle'

program/rust/src/tests/test_full_publisher_set.rs
drozdziak1: I dug a bit deeper. Zero slot warps fails to aggregate, but a single slot warp from 1 to 2 works, followed by a single extra update (my dev debug message shows that aggregate got bumped for that last update). I still managed to simplify the slot stuff with that. Thanks!

program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -0,0 +1,92 @@
...
+        signer::Signer,
+    },
+};
+
+// Verify that the whole publisher set participates in aggregate
+// calculation. This is important for verifying that extra
+// publisher slots on Pythnet are working. Here's how this works:
+//
+// * Fill all publisher slots on a price
+// * Divide the price component array into three parts roughly the
```
guibescos, line 24: This comment became outdated now that you're halving

program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -0,0 +1,92 @@
...
+// same size: first_third, middle_third and last_third
+// * Publish two distinct price values to first/mid third
+// * Verify that the aggregate averages out to an expected value in the middle
+// * Publish PC_STATUS_UNKNOWN prices to first_third to make sure it does not affect later aggregates
+// * Repeat the two-price-value publishing step, using completely
+// different values, this time using mid_third and last_third without first_third
+// * Verify again for the new values meeting in the middle as aggregate.
+#[tokio::test]
+async fn test_full_publisher_set_two_thirds() -> Result<(), Box<dyn std::error::Error>> {
```
guibescos, line 33: The name of this test is outdated too

guibescos (dismissed):
This is very cool. Very happy you managed to do it with only one warp_to_slot. 
I've some nits about the documentation and the name of the test but I'm gonna approve.

guibescos (approved):

PR #388, opened 2023-11-17
Oracle: Make permissions account mandatory for all privileged instructions

drozdziak1:
# Motivation
During an audit of oracle code, Amin and I discovered that it is still possible to circumvent the new permissions PDA mechanism if you simply don't pass the permissions account. If you as an attacker have control over price/product/mapping private keys, you can use the legacy privkey-based access control, which means the oracle has a significantly larger attack surface than necessary. This change makes the permissions account mandatory and adjusts all affected unit tests.

# Summary of changes
* `program/rust/src/tests/pyth_simulator.rs` - always use permissions account in instruction wrappers
* `program/rust/src/utils.rs` - Return error on missing permissions account
* `program/rust/src/tests/` - Add `permissions_account_setup` harness assignments to all affected tests.

# Review Highlights
* `program/rust/src/tests/` - Some of the assertions were removed, as they were testing the legacy privkey access control which is no longer available. That said, it's important that useful assertions were not removed.

### DIFF
--- a/program/rust/src/tests/test_add_product.rs
+++ b/program/rust/src/tests/test_add_product.rs
@@ -5,6 +5,7 @@ use {
             clear_account,
             create_pc_str_t,
             MappingAccount,
+            PermissionAccount,
             ProductAccount,
             PythAccount,
         },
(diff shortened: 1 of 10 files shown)


program/rust/src/tests/test_add_product.rs
```diff
@@ -57,13 +58,25 @@ fn test_add_product() {
...
+    let mut permissions_setup = AccountSetup::new_permission(&program_id);
+    let permissions_account = permissions_setup.as_account_info();
+
+    {
+        let mut permissions_account_data =
+            PermissionAccount::initialize(&permissions_account, PC_VERSION).unwrap();
+        permissions_account_data.master_authority = *funding_account.key;
+        permissions_account_data.data_curation_authority = *funding_account.key;
+        permissions_account_data.security_authority = *funding_account.key;
+    }
```
jayantk, line 70: can you test that the old way of invoking these instructions fails now?

jayantk: err also would be good to test the specific authority that is allowed to do these operations. So maybe set that the other authorities are different

(unless this is already tested elsewhere, which i think it probably is)

jayantk (approved):
I think it would be good to validate that forgetting the permissions account is always a failure now, but lgtm aside from that.

jayantk (approved):
🙇

PR #391, opened 2024-02-15
add support to configure max latency

cctdaniel:
add configurable max latency in slots between send and receive

for price feeds with more publishers this can be reduced instead of using the default 25 value to reduce latency

### DIFF
--- a/program/c/src/oracle/oracle.h
+++ b/program/c/src/oracle/oracle.h
@@ -195,8 +195,9 @@ typedef struct pc_price
   pc_ema_t        twac_;              // time-weighted average conf interval
   int64_t         timestamp_;         // unix timestamp of aggregate price
   uint8_t         min_pub_;           // min publishers for valid price
-  int8_t          drv2_;              // space for future derived values
-  int16_t         drv3_;              // space for future derived values
+  int8_t          message_sent_;      // flag to indicate if the current aggregate price has been sent as a message to the message buffer, 0 if not sent, 1 if sent
+  uint8_t         max_latency_;       // configurable max latency in slots between send and receive
+  int8_t          drv3_;              // space for future derived values
   int32_t         drv4_;              // space for future derived values
   pc_pub_key_t    prod_;              // product id/ref-account
   pc_pub_key_t    next_;              // next price account in list
--- a/program/c/src/oracle/upd_aggregate.h
+++ b/program/c/src/oracle/upd_aggregate.h
@@ -175,7 +175,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
            // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
            // These checks ensure that price - conf and price + conf do not overflow.
            (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) &&
-           slot_diff >= 0 && slot_diff <= PC_MAX_SEND_LATENCY ) {
+           slot_diff <= (ptr->max_latency_ ? ptr->max_latency_ : PC_MAX_SEND_LATENCY) ) {
         numv += 1;
         prcs[ nprcs++ ] = price - conf;
         prcs[ nprcs++ ] = price;
--- a/program/c/src/oracle/upd_aggregate.h
+++ b/program/c/src/oracle/upd_aggregate.h
@@ -175,6 +175,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
            // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
            // These checks ensure that price - conf and price + conf do not overflow.
            (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) &&
+           // slot_diff is implicitly >= 0 due to the check in Rust code ensuring publishing_slot is always less than or equal to the current slot.
            slot_diff <= (ptr->max_latency_ ? ptr->max_latency_ : PC_MAX_SEND_LATENCY) ) {
         numv += 1;
         prcs[ nprcs++ ] = price - conf;
--- a/program/c/src/oracle/upd_aggregate.h
+++ b/program/c/src/oracle/upd_aggregate.h
@@ -171,12 +171,13 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
       int64_t slot_diff = ( int64_t )slot - ( int64_t )( iptr->agg_.pub_slot_ );
       int64_t price     = iptr->agg_.price_;
       int64_t conf      = ( int64_t )( iptr->agg_.conf_ );
+      int64_t max_latency = ptr->max_latency_ ? ptr->max_latency_ : PC_MAX_SEND_LATENCY;
       if ( iptr->agg_.status_ == PC_STATUS_TRADING &&
            // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
            // These checks ensure that price - conf and price + conf do not overflow.
            (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) &&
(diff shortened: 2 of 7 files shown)


program/c/src/oracle/oracle.h
```diff
@@ -195,8 +195,9 @@ typedef struct pc_price
   pc_ema_t        twac_;              // time-weighted average conf interval
   int64_t         timestamp_;         // unix timestamp of aggregate price
   uint8_t         min_pub_;           // min publishers for valid price
-  int8_t          drv2_;              // space for future derived values
-  int16_t         drv3_;              // space for future derived values
+  int8_t          message_sent_;      // flag to indicate if the current aggregate price has been sent as a message to the message buffer, 0 if not sent, 1 if sent
```
cctdaniel, line 198: drive-by: this was already changed in the rust program previously but wasn't changed here

program/c/src/oracle/oracle.h
```diff
@@ -195,8 +195,9 @@ typedef struct pc_price
...
+  uint8_t         max_latency_;       // configurable max latency in slots between send and receive
```
cctdaniel, line 199: we only need `uint8_t` for `max_latency_` so splitting it to `uint8_t` and `int8_t`

program/rust/src/tests/test_twap.rs
```diff
@@ -19,20 +19,24 @@ use {
...
         DataEvent {
-            slot_gap: u64::from(u8::arbitrary(g)) + 1, /* Slot gap is always > 1, because there
-                                                        * has been a succesful aggregation */
-            price:    i64::arbitrary(g),
-            conf:     u64::arbitrary(g),
+            slot_gap:    u64::from(u8::arbitrary(g)) + 1, /* Slot gap is always > 1, because there
+                                                           * has been a succesful aggregation */
+            price:       i64::arbitrary(g),
+            conf:        u64::arbitrary(g),
+            max_latency: 0,
```
guibescos, line 39: I think here you want something abritary to extend the test coverage

program/rust/src/tests/test_twap.rs
```diff
@@ -87,12 +96,18 @@ impl PriceCumulative {
...
     pub fn check_num_down_slots(&self, data: &[DataEvent]) {
         assert_eq!(
-            data.iter()
-                .fold(0, |acc, x| if x.slot_gap > PC_MAX_SEND_LATENCY.into() {
-                    acc + (x.slot_gap - PC_MAX_SEND_LATENCY as u64)
+            data.iter().fold(0, |acc, x| {
+                let latency_threshold = if x.max_latency == 0 {
+                    PC_MAX_SEND_LATENCY.into()
+                } else {
+                    x.max_latency as u64
```
guibescos, line 103: `.into()` seems cleaner, also I don't think this pathway is ever explored because DataEvent has max_latency equal to 0 always

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -269,6 +286,133 @@ fn test_upd_aggregate() {
         assert_eq!(price_data.prev_conf_, 85);
         assert_eq!(price_data.prev_timestamp_, 5);
     }
+
+    // check max_latency_ = 0 defaults to PC_MAX_SEND_LATENCY
```
guibescos, line 290: mm arguably this is already checked above?

program/c/src/oracle/upd_aggregate.h
```diff
@@ -175,7 +175,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
            // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
            // These checks ensure that price - conf and price + conf do not overflow.
            (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) &&
-           slot_diff >= 0 && slot_diff <= PC_MAX_SEND_LATENCY ) {
```
guibescos, line 178: Was it voluntary to remove` slot_diff >= 0`

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -269,6 +286,133 @@ fn test_upd_aggregate() {
...
+        assert_eq!(price_data.prev_price_, 245);
+        assert_eq!(price_data.prev_conf_, 85);
+        assert_eq!(price_data.prev_timestamp_, 5);
+    }
+
+    // ensure the update occurs within the PC_MAX_SEND_LATENCY limit of 25 slots, allowing the aggregated price to reflect both p5 and p6 contributions
+    {
+        let mut price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        price_data.num_ = 2;
+        price_data.last_slot_ = 1025;
```
guibescos, line 326: I think this would be more readable if you don't update `last_slot` and `agg_.pub_slot` and you keep them equal to 1000 like above.
To create situations where one of the prices expires you just need to tune the `pub_slot_` parameter of the price component.
Here you're modifying two parameters at a time so it's a bit confusing

program/c/src/oracle/upd_aggregate.h
ali-bahjati: I think we need to keep it to encourage a publisher sending a price in future slot and get that price included for all slots.

program/rust/src/tests/test_twap.rs
```diff
@@ -163,7 +198,12 @@ fn test_twap_unit() {
     );
     assert_eq!(
         price_cumulative_overflow.num_down_slots,
-        u64::MAX - PC_MAX_SEND_LATENCY as u64
+        u64::MAX
+            - if TEST_MAX_LATENCY == 0 {
```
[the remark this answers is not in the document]
ali-bahjati: generally I recommend adding different test cases measuring different values for max_latency in TWAP including 0.

ali-bahjati (commented):
Very nice! I will look again after you addressed the feedbacks (just to be extra careful).

program/c/src/oracle/upd_aggregate.h
cctdaniel: this is redundant, there is a check slightly above to return false if `agg_.pub_slot_` is greater than current slot
```
  // only re-compute aggregate in next slot
  if ( slot <= ptr->agg_.pub_slot_ ) {
    return false;
  }
```

guibescos: ok

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -269,6 +279,134 @@ fn test_upd_aggregate() {
...
+        ));
+    }
+
+    {
+        let price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+
+        assert_eq!(price_data.max_latency_, 0);
+        assert_eq!(price_data.agg_.price_, 445);
+        assert_eq!(price_data.agg_.conf_, 55);
+        assert_eq!(price_data.twap_.val_, 168);
```
guibescos, line 339: I'm a bit confused about these twaps, imo something that can't be checked manually easily shouldn't be here.

guibescos (dismissed):

program/rust/src/tests/test_upd_aggregate.rs
cctdaniel: are you suggesting to remove the twap and twac checks here?

guibescos (dismissed):

program/c/src/oracle/upd_aggregate.h
```diff
@@ -175,7 +175,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
...
+           slot_diff <= (ptr->max_latency_ ? ptr->max_latency_ : PC_MAX_SEND_LATENCY) ) {
```
ali-bahjati, line 178: I recommend setting the max latency as a variable outside this if (just for more code clarity)

ali-bahjati (approved):
Nice! I left some minor comments.

ali-bahjati (dismissed):
Nice! I left some minor comments.

ali-bahjati (approved):

PR #394, opened 2024-02-21
feat: aggregate in the same slot

cctdaniel:
update aggregation logic to aggregate in the same slot, also change ema and twap to calculate on the same slot


program/c/src/oracle/upd_aggregate.h
```diff
@@ -174,8 +176,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
       if ( iptr->agg_.status_ == PC_STATUS_TRADING &&
            // No overflow for INT64_MIN+conf or INT64_MAX-conf as 0 < conf < INT64_MAX
            // These checks ensure that price - conf and price + conf do not overflow.
-           (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) &&
-           slot_diff >= 0 && slot_diff <= PC_MAX_SEND_LATENCY ) {
+           (int64_t)0 < conf && (INT64_MIN + conf) <= price && price <= (INT64_MAX-conf) && slot_diff <= PC_MAX_SEND_LATENCY ) {
```
cctdaniel, line 179: removed `slot_diff >= 0` here because there is a check above already that makes sure slot_diff is >= 0
```
  if ( slot < ptr->agg_.pub_slot_ ) {
    return false;
  }
  ```

File: program/c/src/oracle/upd_aggregate.h [hunk 2]
```diff
@@ -222,7 +223,9 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
   ptr->agg_.price_  = agg_price;
   ptr->agg_.conf_   = (uint64_t)agg_conf;
 
-  upd_twap( ptr, agg_diff );
+  if ( update_twap ) {
+    upd_twap( ptr, agg_diff );
+  }
   return true;
```
guibescos, line 229: This return true triggers and accumulator message and a PriceCumulative update, you need to be careful about that

program/rust/src/processor/upd_price.rs
```diff
@@ -175,21 +176,82 @@ pub fn upd_price(
...
 
-    // Try to update the aggregate
-    #[allow(unused_variables)]
-    let mut aggregate_updated = false;
+    {
+        // Reload price data as a struct after c_upd_aggregate() borrow is dropped
+        let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+
+        // Try to update the publisher's price
+        if is_component_update(cmd_args)? {
```
guibescos, line 184: mmmmmm moving this up here seems complicated, maybe this should be a different PR?

guibescos: I'm worried about the side effects of this

program/rust/src/processor/upd_price.rs
```diff
@@ -175,21 +176,82 @@ pub fn upd_price(
...
+                price_data.prev_price_cumulative = price_data.price_cumulative;
+            }
+            price_data.twap_ = price_data.prev_twap_;
+            price_data.twac_ = price_data.prev_twac_;
+            price_data.price_cumulative = price_data.prev_price_cumulative;
+        }
+        unsafe {
+            c_upd_twap(
+                price_account.try_borrow_mut_data()?.as_mut_ptr(),
+                slots_since_last_update as i64, // Ensure slots_since_last_update is cast to i64, as expected by the function signature
```
guibescos, line 247: Does this break ema on Solana mainnet?

program/rust/src/processor/upd_price.rs
```diff
@@ -144,6 +144,7 @@ pub fn upd_price(
     let mut publisher_index: usize = 0;
     let latest_aggregate_price: PriceInfo;
 
+    #[cfg(feature = "pythnet")]
```
guibescos, line 147: Wait why is this only pythnet

program/rust/src/processor/upd_price.rs
```diff
@@ -175,21 +176,82 @@ pub fn upd_price(
         )?;
     }
 
-    // Try to update the aggregate
-    #[allow(unused_variables)]
-    let mut aggregate_updated = false;
+    {
```
guibescos, line 179: Does this component update happen twice on mainnet solana

program/rust/src/processor/upd_price.rs
```diff
@@ -175,21 +176,82 @@ pub fn upd_price(
...
+        }
+        unsafe {
+            c_upd_twap(
+                price_account.try_borrow_mut_data()?.as_mut_ptr(),
+                slots_since_last_update as i64, // Ensure slots_since_last_update is cast to i64, as expected by the function signature
             );
+            #[cfg(feature = "pythnet")]
+            {
+                let mut price_data =
+                    load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
```
guibescos, line 252: Maybe you can move this to the block that starts in line 229 so it's more readable

program/rust/src/processor/upd_price.rs
```diff
@@ -175,21 +176,82 @@ pub fn upd_price(
...
+            // get price data here
+            let mut price_data =
+                load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+            // check if its the first price update in the slot
+            if slots_since_last_update > 0 {
+                price_data.prev_twap_ = price_data.twap_;
+                price_data.prev_twac_ = price_data.twac_;
+                price_data.prev_price_cumulative = price_data.price_cumulative;
+            }
+            price_data.twap_ = price_data.prev_twap_;
```
guibescos, line 240: What happens the first time this codepath is reached if `price_data.prev_twap_` and `price_data.prev_price_cumulative` are uninitialized

program/c/src/oracle/upd_aggregate.h
```diff
@@ -134,14 +134,6 @@ static inline void upd_twap(
...
-  if ( slot <= ptr->agg_.pub_slot_ ) {
-    return false;
-  }
-
-  // get number of slots from last published valid price
-  int64_t agg_diff = ( int64_t )slot - ( int64_t )( ptr->last_slot_ );
-
   // Update the value of the previous price, if it had TRADING status.
   if ( ptr->agg_.status_ == PC_STATUS_TRADING ) {
     ptr->prev_slot_      = ptr->agg_.pub_slot_;
```
guibescos, line 139: I feel like this logic gets broken since in the past the previous aggregate was supposed to be from a previous slot

program/rust/src/processor/upd_price.rs
cctdaniel: yeah this should be for the chunk below, good catch

program/rust/src/processor/upd_price.rs
cctdaniel: when I ran the tests it seems like it's initialized to 0 by rust unless I am wrong?

program/c/src/oracle/upd_aggregate.h
cctdaniel: yeah you're right, moved this to `upd_price.rs` just before it calls `c_upd_aggregate`

program/rust/src/processor/upd_price.rs
```diff
@@ -175,20 +176,85 @@ pub fn upd_price(
...
+                publisher_price.price_ = cmd_args.price;
+                publisher_price.conf_ = cmd_args.confidence;
+                publisher_price.status_ = status;
+                publisher_price.pub_slot_ = cmd_args.publishing_slot;
+            }
+        }
+    }
+
+    let should_call_c_upd_aggregate: bool = if cfg!(feature = "pythnet") {
+        clock.slot >= latest_aggregate_price.pub_slot_
```
ali-bahjati, line 206: this is always true right?

cctdaniel:
not sure why it says `112 passed`, running locally gives me `121 passed` 🤔

ali-bahjati (commented):
The change looks good to me. Everyone should double check the side-effects of all other functions as well as the diff shown in the PR.

I have some suggestions:
- can we rename `upd_twap` to a better name because that's ema?
- i'm wondering whether it gets cleaner if `upd_twap` and `update_price_cumulative` always read from prev fields and write on the current fields inside themself (so we don't hack them each time by writing `current = previous`)

program/rust/src/processor/upd_price.rs
guibescos: The prolbem is then  price_data.twap_ will get assigned to 0 too

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +162,76 @@ pub fn upd_price(
...
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.prev_slot_
+                as i64;
+        {
             let mut price_data =
                 load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+            // check if its the first price update in the slot
+            if slots_since_last_update > 0 {
+                price_data.prev_twap_ = price_data.twap_;
+                price_data.prev_twac_ = price_data.twac_;
+                price_data.prev_price_cumulative = price_data.price_cumulative;
```
guibescos, line 221: It might be more readable if you move the assignments for `prev_twap_`, `prev_twac_`, `prev_price_cumulative` to the block where you assign `prev_slot_`, `prev_price_ `etc..

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +162,76 @@ pub fn upd_price(
...
+            }
+            price_data.twap_ = price_data.prev_twap_;
+            price_data.twac_ = price_data.prev_twac_;
+            price_data.price_cumulative = price_data.prev_price_cumulative;
+            price_data.update_price_cumulative()?;
             // We want to send a message every time the aggregate price updates. However, during the migration,
             // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
             // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
             // will send the message.
             price_data.message_sent_ = 0;
```
guibescos, line 231: (maybe future PR) I'm pretty sure we don't need `message_sent_` anymore, since now everytime a publisher publishes we're gonna attempt an aggregation and attempt to send

program/rust/src/accounts/price.rs
```diff
@@ -31,62 +31,113 @@ mod price_pythnet {
...
+        pub num_:                  u32,
         /// Number of valid quotes for the last aggregation
-        pub num_qt_:            u32,
+        pub num_qt_:               u32,
         /// Last slot with a succesful aggregation (status : TRADING)
-        pub last_slot_:         u64,
+        pub last_slot_:            u64,
         /// Second to last slot where aggregation was attempted
-        pub valid_slot_:        u64,
+        pub valid_slot_:           u64,
```
guibescos, line 100: This `valid_slot_` field doesn't make that much sense anymore after the change

guibescos (changes requested):
IMO there's a critical bug here. 
prev_twap_, prev_twac etc... will get initialized to 0, so there's a risk that twap_, twac_ become 0 when the change gets deployed

program/rust/src/tests/test_ema.rs
```diff
@@ -1,23 +1,133 @@
...
+
+    {
+        let price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        assert_eq!(price_data.prev_twap_.val_, 0);
+        assert_eq!(price_data.prev_twac_.val_, 0);
+        assert_eq!(price_data.twap_.val_, 10);
+        assert_eq!(price_data.twac_.val_, 1);
+    }
+
+    // add new test for multiple publishers and ensure that price_cumulative is updated correctly
```
guibescos, line 89: comment is wrong, we're not checking `price_cumulative`

guibescos: I don't get exactly what this test is testing wrt to this PR

program/rust/src/tests/test_publish.rs
```diff
@@ -64,21 +64,6 @@ async fn test_publish() {
...
     .unwrap();
 
-    {
-        let price_data = sim
-            .get_account_data_as::<PriceAccount>(price)
-            .await
-            .unwrap();
-
-        assert_eq!(price_data.comp_[0].latest_.price_, 150);
-        assert_eq!(price_data.comp_[0].latest_.conf_, 7);
```
guibescos, line 74: Why remove this?

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -78,13 +77,38 @@ fn test_upd_aggregate() {
...
+        price_data.last_slot_ = 1000;
+        price_data.agg_.pub_slot_ = 1000;
+        price_data.comp_[0].latest_ = p1;
+    }
+    unsafe {
+        assert!(c_upd_aggregate(
+            price_account.try_borrow_mut_data().unwrap().as_mut_ptr(),
+            1000,
+            1,
+        ));
```
guibescos, line 93: Did you mean to add some asserts after this?

program/rust/src/tests/test_upd_price.rs
```diff
@@ -106,8 +106,8 @@ fn test_upd_price() {
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
         assert_eq!(price_data.valid_slot_, 0);
         assert_eq!(price_data.agg_.pub_slot_, 1);
-        assert_eq!(price_data.agg_.price_, 0);
-        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.agg_.price_, 42);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
     }
 
     // add next price in new slot triggering snapshot and aggregate calc
```
guibescos, line 113: Outdate comment

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -134,10 +158,6 @@ fn test_upd_aggregate() {
         assert_eq!(price_data.agg_.conf_, 55);
         assert_eq!(price_data.num_qt_, 2);
         assert_eq!(price_data.timestamp_, 2);
-        assert_eq!(price_data.prev_slot_, 1000);
```
jayantk, line 137: I understand why you've deleted these (because upd_aggregate no longer updates them). However, after making this change, we no longer have a test for the prev_price updates.

I think you need to add an equivalent test back somewhere else, probably in test_upd_price or test_upd_price_v2

program/rust/src/tests/test_upd_price.rs
```diff
@@ -248,7 +248,7 @@ fn test_upd_price() {
         assert_eq!(price_data.valid_slot_, 5);
         assert_eq!(price_data.agg_.pub_slot_, 6);
         assert_eq!(price_data.agg_.price_, 81);
-        assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN); // in pythnet aggregation is on the same slot so status turns to unknown
```
jayantk, line 251: why in pythnet? this whole codebase is now pythnet only right?

jayantk (changes requested):
I think this is a great start. There are a couple of subtle bugs though so please fix those and we will look again.

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +162,76 @@ pub fn upd_price(
...
+    // If the aggregate was successfully updated, calculate the difference and update TWAP.
+    if updated {
+        let agg_diff = (clock.slot as i64)
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.prev_slot_
+                as i64;
+        {
             let mut price_data =
                 load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+            // check if its the first price update in the slot
+            if slots_since_last_update > 0 {
```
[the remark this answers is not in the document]
guibescos: I'm not sure I agree about this one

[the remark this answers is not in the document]
cctdaniel: > I think this if statement is wrong. What if you have a slot where the first aggregate update sets the status to UNKNOWN, then a later update sets it to TRADING? In that case, this if statement will return false and we will forget to update the twap.
> 
> I think the right check here is to explicitly compare the clock slot with a stored last updated slot for the prev_twap_ (etc) fields.

I'm not sure I get this, `c_upd_aggregate` will only return `True` if agg price is successfully updated and `ptr->agg_.status_` will be `PC_STATUS_TRADING`, and in that case this block will only be ran if `ptr->agg_.status_ = PC_STATUS_TRADING`

program/rust/src/tests/test_ema.rs
cctdaniel: fixed the comment, can you give more contexts as to which specific part is unclear? this is to test that multiple publishers with price updates should lead to ema being updated

program/rust/src/tests/test_upd_price.rs
cctdaniel: will remove to prevent confusion

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +162,76 @@ pub fn upd_price(
...
+                load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+            let publisher_price = &mut price_data.comp_[publisher_index].latest_;
+            publisher_price.price_ = cmd_args.price;
+            publisher_price.conf_ = cmd_args.confidence;
+            publisher_price.status_ = status;
+            publisher_price.pub_slot_ = cmd_args.publishing_slot;
+        }
+    }
+
+    // get number of slots from last update
```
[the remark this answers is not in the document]
cctdaniel: correct me if I'm wrong but when `price_data.agg_` is assigned to `latest_aggregate_price` on couple of lines above, and given that `price_data.agg_` is of type `PriceInfo` which implements the `Copy` trait, I think when we do the assignment, `latest_aggregate_price` holds its own copy of the data which is separate from `price_data.agg_` so any subsequent changes to `price_data.agg_` shouldn't affect `latest_aggregate_price`

program/rust/src/processor/upd_price.rs
cctdaniel: hmm the reason why I moved it here is because it requires the aggregate to be updated before `twap`,`twac`,`prev_price_cumulative`, if I moved it up before `c_upd_aggregate` then this might be updated even if the aggregate didn't succeed

program/rust/src/accounts/price.rs
cctdaniel: yeah I think we can do it in a separate PR, what do you think?

program/rust/src/tests/test_publish.rs
cctdaniel: added test back

program/rust/src/tests/test_upd_aggregate.rs
cctdaniel: yes 🤦‍♂️

program/rust/src/processor/upd_price.rs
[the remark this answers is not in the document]
cctdaniel: changed slots_since_last_update to be `clock.slot - price_data.last_slot_` instead since `last_slot` will only be updated if the agg price is successfully updated

program/rust/src/processor/upd_price.rs
cctdaniel: added a check to see if the program upgrade has happened in the current slot and aggregate price has been updated, if so, use the old logic to update twap/twac/price_cumulative

cctdaniel:
> The change looks good to me. Everyone should double check the side-effects of all other functions as well as the diff shown in the PR.
> 
> I have some suggestions:
> 
> * can we rename `upd_twap` to a better name because that's ema?
> * i'm wondering whether it gets cleaner if `upd_twap` and `update_price_cumulative` always read from prev fields and write on the current fields inside themself (so we don't hack them each time by writing `current = previous`)

agreed with the above points, happy to do this in a separate PR so that it wont further bloat this PR

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +169,81 @@ pub fn upd_price(
...
+            price_account.try_borrow_mut_data()?.as_mut_ptr(),
+            clock.slot,
+            clock.unix_timestamp,
+        )
+    };
+
+    // If the aggregate was successfully updated, calculate the difference and update TWAP.
+    if updated {
+        let agg_diff = (clock.slot as i64)
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.prev_slot_
```
guibescos, line 213: IMO you shouldn't aggregate IF 
`(price_data.prev_twap_.denom_ == 0 && slots_since_last_update == 0)` 

Otherwise you aggregate twice on the slot where the program gets upgraded

program/rust/src/processor/upd_price.rs
```diff
@@ -161,37 +169,81 @@ pub fn upd_price(
...
+            price_data.prev_conf_ = price_data.agg_.conf_;
+            price_data.prev_timestamp_ = clock.unix_timestamp;
+        }
+    }
+
+    let updated = unsafe {
+        // NOTE: c_upd_aggregate must use a raw pointer to price
+        // data. Solana's `<account>.borrow_*` methods require exclusive
+        // access, i.e. no other borrow can exist for the account.
+        c_upd_aggregate(
```
guibescos, line 203: IMO you shouldn't aggregate IF 
`(price_data.prev_twap_.denom_ == 0 && slots_since_last_update == 0)` 

Otherwise you aggregate twice on the slot where the program gets upgraded. But only the first aggregate makes it into the twaps.

program/rust/src/tests/test_ema.rs
```diff
@@ -1,23 +1,190 @@
...
+    process_instruction(
+        &program_id,
+        &[
+            funding_account.clone(),
+            price_account.clone(),
+            clock_account.clone(),
+        ],
+        &instruction_data,
+    )?;
+
```
guibescos, line 112: Why no assert between these two instructions?

program/rust/src/tests/test_upd_price.rs
```diff
@@ -248,7 +272,11 @@ fn test_upd_price() {
...
         assert_eq!(price_data.agg_.price_, 81);
-        assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.prev_slot_, 5);
+        assert_eq!(price_data.prev_price_, 81);
+        assert_eq!(price_data.prev_conf_, 2);
+        assert_eq!(price_data.prev_timestamp_, 0);
     }
 
     // Crank one more time and aggregate should be unknown
```
guibescos, line 282: `// Crank one more time and aggregate should be unknown` comments now are weird since we don't need to crank the aggregation

program/rust/src/tests/test_ema.rs
cctdaniel: this was mainly ensuring that 2 consec updates should update ema correctly hence the single assert at the end, but will add another block here to assert the first price update

program/rust/src/tests/test_upd_price.rs
cctdaniel: removed the whole assert blocks for "crank again" comments

program/rust/src/processor/upd_price.rs
```diff
@@ -173,25 +216,42 @@ pub fn upd_price(
...
+    // If the aggregate was successfully updated, calculate the difference and update TWAP.
+    if updated {
+        let agg_diff = (clock.slot as i64)
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.prev_slot_
+                as i64;
+        {
             let mut price_data =
                 load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
+
+            if !(noninitial_price_update_after_program_upgrade) {
```
jayantk, line 231: This logic is very confusing and I'm not at all sure that it's correct.

(As a general code clarity thing, I think you should *never* name a variable as a negative and always name them in the positive. E.g., "noninitial" is bad. "initial_price_update" is good.)

jayantk: I think the thing you want to do is have a variable which is like `initial_price_update_after_program_upgrade` and then on line 221 write something like

```
if (initial_price_update_after_program_upgrade) {
                   price_data.prev_twap_ = price_data.twap_;
                   price_data.prev_twac_ = price_data.twac_;
                   price_data.prev_price_cumulative = price_data.price_cumulative;
}
```

Basically detect the exact time when the variables haven't been initialized, then initialize them to what you expect so that the (new) standard twap / price cumulative update logic works as expected.

program/rust/src/processor/upd_price.rs
```diff
@@ -173,25 +216,42 @@ pub fn upd_price(
...
+            }
+            price_data.update_price_cumulative()?;
             // We want to send a message every time the aggregate price updates. However, during the migration,
             // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
             // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
             // will send the message.
             price_data.message_sent_ = 0;
-            price_data.update_price_cumulative()?;
+        }
+        unsafe {
```
jayantk, line 253: Can you please move the computation of agg_diff and the setting of the needed twap_ variables to above this line so that the relationship between these pieces of the code is clear to the reader?

jayantk (commented):
I looked through part of it and will continue on in a bit. I think the logic in the price update needs to be simplified (and can be simplified) because it's not at all clear that it's correct right now. See inline suggestion.

program/rust/src/processor/upd_price.rs
jayantk: maybe you don't even need the initial_price_update part and can literally go with 'program_upgraded`

program/rust/src/tests/test_upd_price_v2.rs
```diff
@@ -141,11 +139,10 @@ fn test_upd_price_v2() -> Result<(), Box<dyn std::error::Error>> {
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
         assert_eq!(price_data.valid_slot_, 1);
         assert_eq!(price_data.agg_.pub_slot_, 3);
-        assert_eq!(price_data.agg_.price_, 42);
+        assert_eq!(price_data.agg_.price_, 81);
         assert_eq!(price_data.agg_.conf_, 2);
         assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
-
-        assert_eq!(price_data.price_cumulative.price, 3 * 42);
+        assert_eq!(price_data.price_cumulative.price, 42 + 2 * 81);
```
jayantk, line 145: this doesn't make sense. I think the right answer is `42 * 2 + 81` (because it was 42 on slot 1, 42 on slot 2, and then 81 on slot 3)

jayantk: hm... it looks like this was the behavior of price cumulative before this PR though. That doesn't make a lot of sense to me though.

program/rust/src/processor/upd_price.rs
```diff
@@ -216,11 +228,7 @@ pub fn upd_price(
             let mut price_data =
                 load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
 
-            // Check if the program upgrade has happened in the current slot and aggregate price has been updated, if so, use the old logic to update twap/twac/price_cumulative.
-            // This is to ensure that twap/twac/price_cumulative are calculated correctly during the migration.
-            // We check if prev_twap_.denom_ is == 0 because when the program upgrade has happened, denom_ is initialized to 0 and it can only stay the same or increase while numer_ can be negative if prices are negative.
-            // And we check if slots_since_last_update == 0 to check if the aggregate price has been updated in the current slot.
-            if !(price_data.prev_twap_.denom_ == 0 && slots_since_last_update == 0) {
+            if !(noninitial_price_update_after_program_upgrade) {
```
guibescos, line 231: Note that you don't need this if since statement since if  noninitial_price_update_after_program_upgrade, updated is false anyway

program/rust/src/processor/upd_price.rs
guibescos: @jayantk This approach doesn't really work. Let's imagine the following situation where the following actions happen in the same slot:
- upd_price
- program upgrade
- upd_price 

First update price updates the aggregate and the twap.
Second update price will create a new aggregate and update twap with the aggregate.
As a result twap has been updated twice.

In the slot where the program update happens, you have no choice but use the old logic because a twap update might already have happened and you have no access to the twap before that update.

cctdaniel: the opposite of `noninitial_price_update_after_program_upgrade` would be `price_data.prev_twap_.denom_ != 0 || slots_since_last_update != 0` which translates to `initial_price_updated_or_program_not_updated` which seemed to be more difficult to understand imo, `program_updated` hides some details for e.g. it has to be the noninitial price updates after program is updated

program/rust/src/tests/test_upd_price.rs
```diff
@@ -106,11 +110,15 @@ fn test_upd_price() {
...
         assert_eq!(price_data.valid_slot_, 0);
         assert_eq!(price_data.agg_.pub_slot_, 1);
-        assert_eq!(price_data.agg_.price_, 0);
-        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.agg_.price_, 42);
+        assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
+        assert_eq!(price_data.prev_slot_, 0);
+        assert_eq!(price_data.prev_price_, 0);
+        assert_eq!(price_data.prev_conf_, 0);
+        assert_eq!(price_data.prev_timestamp_, 0);
```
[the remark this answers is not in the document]
cctdaniel: added

program/rust/src/tests/test_upd_price_v2.rs
cctdaniel: yeah this was the existing behavior, we can change this if we want to but imo we should do this on a separate PR

program/rust/src/tests/test_ema.rs
```diff
@@ -1,23 +1,198 @@
...
+    {
```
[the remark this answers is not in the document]
cctdaniel: agreed with what you said, added a new helper method `add_publisher` but for `update_publisher_price` I don't think abstracting it is going to make it any easier to understand since it's already pretty straightforward, I think the reason its hard to understand here is because the first `populate_instruction()` is pretty high above, will move it below right before `process_instruction`

program/rust/src/tests/test_ema.rs
```diff
@@ -1,23 +1,198 @@
...
+        let price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        assert_eq!(price_data.prev_twap_.val_, 0);
+        assert_eq!(price_data.prev_twac_.val_, 0);
+        assert_eq!(price_data.twap_.val_, 10);
+        assert_eq!(price_data.twac_.val_, 1);
+    }
```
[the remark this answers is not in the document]
cctdaniel: yes but more specifically its because the `prev_twap.denom_` is 0, will add the assertion for that and add a comment to clarify

program/rust/src/tests/test_ema.rs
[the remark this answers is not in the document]
cctdaniel: also for `update_publisher_price` this is existing code, and we can refactor them in a following PR, what do you think?

program/rust/src/processor/upd_price.rs
```diff
@@ -200,7 +200,7 @@ pub fn upd_price(
             price_data.prev_slot_ = price_data.agg_.pub_slot_;
             price_data.prev_price_ = price_data.agg_.price_;
             price_data.prev_conf_ = price_data.agg_.conf_;
-            price_data.prev_timestamp_ = clock.unix_timestamp;
+            price_data.prev_timestamp_ = price_data.timestamp_;
```
guibescos, line 203: wow did you catch a bug that was in the old code here?

cctdaniel: no this is introduced in this pr

program/rust/src/tests/test_ema.rs
```diff
@@ -1,23 +1,191 @@
...
+            clock_account.clone(),
+        ],
+        &instruction_data,
+    )?;
+
+    {
+        let price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        assert_eq!(price_data.prev_twap_.val_, 10);
+        assert_eq!(price_data.prev_twac_.val_, 1);
+        assert_eq!(price_data.twap_.val_, 12);
```
jayantk, line 129: why is this value 12? I get why the one above is 15 (10 on the first slot, 20 on the second slot). But then adding a price of 30 seems like it should make this value go up, right? 

I.e., how does the EMA decrease when the aggregate price in the slot increases?

program/rust/src/processor/upd_price.rs
```diff
@@ -149,6 +152,19 @@ pub fn upd_price(
...
+        // Check if the program upgrade has happened in the current slot and aggregate price has been updated, if so, use the old logic to update twap/twac/price_cumulative.
+        // This is to ensure that twap/twac/price_cumulative are calculated correctly during the migration.
+        // We check if prev_twap_.denom_ is == 0 because when the program upgrade has happened, denom_ is initialized to 0 and it can only stay the same or increase while numer_ can be negative if prices are negative.
+        // And we check if slots_since_last_update == 0 to check if the aggregate price has been updated in the current slot.
+        noninitial_price_update_after_program_upgrade =
+            price_data.prev_twap_.denom_ == 0 && slots_since_last_update == 0;
+
+        // We assign the current aggregate price to latest_aggregate_price before calling c_upd_aggregate because
+        // c_upd_aggregate directly modifies the memory of price_data.agg_ to update its values. This is crucial for
+        // comparisons or operations that rely on the aggregate price state before the update such as slots_since_last_update.
```
jayantk, line 167: I don't think this comment is right. The code below never uses latest_aggregate_price *after* the call to c_upd_aggregate

program/rust/src/tests/test_ema.rs
ali-bahjati: the reason for a lower number is the increased confidence. it seems that a new price has the weight of 1/confidence when being added. The formula seems complex and i’m not sure i get it. But if you just add a single price by a publisher with agg 25 and conf 5 then the result will be the same.

Part of the forumla that has it:
```
numer = numer * decay + val * 1/conf
denom = denum * decay + 1/conf
=> val = numer / denom 
```

I think it is worth documenting the reason.

ali-bahjati (approved):

jayantk (approved):

ali-bahjati (dismissed):

program/rust/src/processor/upd_price.rs
```diff
@@ -128,7 +128,8 @@ pub fn upd_price(
     let clock = Clock::from_account_info(clock_account)?;
 
     let mut publisher_index: usize = 0;
-    let latest_aggregate_price: PriceInfo;
+    let slots_since_last_update: u64;
```
guibescos, line 131: nit: rename to `slots_since_last_successful_aggregate` ?

program/rust/src/processor/upd_price.rs
```diff
@@ -173,25 +210,41 @@ pub fn upd_price(
...
+            price_data.twap_ = price_data.prev_twap_;
+            price_data.twac_ = price_data.prev_twac_;
+            price_data.price_cumulative = price_data.prev_price_cumulative;
+
+            price_data.update_price_cumulative()?;
             // We want to send a message every time the aggregate price updates. However, during the migration,
             // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
             // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
             // will send the message.
             price_data.message_sent_ = 0;
```
guibescos, line 241: Perhaps future refactor, get rid of `message_sent_` since now we should send it to the message-buffer everytime (since there's a new update everytime).

program/c/src/oracle/upd_aggregate.h
```diff
@@ -134,14 +134,6 @@ static inline void upd_twap(
...
-  // Update the value of the previous price, if it had TRADING status.
-  if ( ptr->agg_.status_ == PC_STATUS_TRADING ) {
-    ptr->prev_slot_      = ptr->agg_.pub_slot_;
-    ptr->prev_price_     = ptr->agg_.price_;
-    ptr->prev_conf_      = ptr->agg_.conf_;
-    ptr->prev_timestamp_ = ptr->timestamp_;
-  }
-
   // update aggregate details ready for next slot
   ptr->valid_slot_ = ptr->agg_.pub_slot_;// valid slot-time of agg. price
```
guibescos, line 138: feel like we should deprecate valid slot here, this field serves no purpose after this PR

program/rust/src/tests/test_ema.rs
```diff
@@ -148,3 +316,19 @@ struct OutputRecord {
...
+    cmd.confidence = conf;
+    cmd.publishing_slot = pub_slot;
+    cmd.unused_ = 0;
+}
+
+fn add_publisher(price_account: &mut AccountInfo, publisher_key: &Pubkey, index: usize) {
+    let mut price_data = load_checked::<PriceAccount>(price_account, PC_VERSION).unwrap();
+    price_data.num_ = (index + 1) as u32;
+    price_data.comp_[index].pub_ = *publisher_key;
+}
```
guibescos, line 334: This function doesn't need to know `index` since you can use the. current `num_` as `index`

program/c/src/oracle/upd_aggregate.h
guibescos: ok I see I had already made this comment haha

program/rust/src/tests/test_publish.rs
```diff
@@ -73,10 +73,9 @@ async fn test_publish() {
         assert_eq!(price_data.comp_[0].latest_.price_, 150);
         assert_eq!(price_data.comp_[0].latest_.conf_, 7);
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
-
-        assert_eq!(price_data.comp_[0].agg_.price_, 0);
-        assert_eq!(price_data.comp_[0].agg_.conf_, 0);
-        assert_eq!(price_data.comp_[0].agg_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.comp_[0].agg_.price_, 150);
```
guibescos, line 76: Note: you can update test_publisher_set.rs to remove the `// Final price update to trigger aggregation` part since we don't need to trigger the aggregation anymore

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -78,13 +77,51 @@ fn test_upd_aggregate() {
...
     // single publisher
     {
         let mut price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
         price_data.num_ = 1;
+        price_data.num_qt_ = 0;
         price_data.last_slot_ = 1000;
+        price_data.timestamp_ = 0;
         price_data.agg_.pub_slot_ = 1000;
         price_data.comp_[0].latest_ = p1;
+        price_data.prev_slot_ = 0;
```
guibescos, line 118: nit: all these fields are 0 already at initialization. I'd remove these assignments

program/rust/src/tests/test_upd_price_v2.rs
```diff
@@ -77,12 +77,11 @@ fn test_upd_price_v2() -> Result<(), Box<dyn std::error::Error>> {
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
         assert_eq!(price_data.valid_slot_, 0);
         assert_eq!(price_data.agg_.pub_slot_, 1);
-        assert_eq!(price_data.agg_.price_, 0);
```
guibescos, line 80: This file is a close of test_upd_price.rs that was meant to be run only for pythnet.
Since we don't have the pythnet feature anymore I think we should merge `test_upd_price.rs` and `test_upd_price_v2.rs`

program/rust/src/tests/test_upd_price.rs
```diff
@@ -78,8 +81,12 @@ fn test_upd_price() {
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
         assert_eq!(price_data.valid_slot_, 0);
         assert_eq!(price_data.agg_.pub_slot_, 1);
-        assert_eq!(price_data.agg_.price_, 0);
-        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+        assert_eq!(price_data.agg_.price_, 42);
```
guibescos, line 84: seems weird not to assert `last_slot_` in this test

program/rust/src/tests/test_upd_price.rs
```diff
@@ -161,6 +177,11 @@ fn test_upd_price() {
         assert_eq!(price_data.agg_.pub_slot_, 4);
         assert_eq!(price_data.agg_.price_, 81);
         assert_eq!(price_data.agg_.status_, PC_STATUS_TRADING);
+        assert_eq!(price_data.prev_slot_, 3);
+        assert_eq!(price_data.prev_price_, 81);
+        assert_eq!(price_data.prev_conf_, 2);
+        assert_eq!(price_data.timestamp_, 4); // We only check for timestamp_ here because test_upd_price doesn't directly update timestamp_, this is updated through c_upd_aggregate which is tested in test_upd_aggregate, but we assert here to show that in subsequent asserts for prev_tim
```
guibescos, line 183: this comment seems unfinished

program/rust/src/tests/test_upd_price_v2.rs
```diff
@@ -77,12 +77,11 @@ fn test_upd_price_v2() -> Result<(), Box<dyn std::error::Error>> {
         assert_eq!(price_data.comp_[0].latest_.status_, PC_STATUS_TRADING);
         assert_eq!(price_data.valid_slot_, 0);
```
guibescos, line 78: In this file we're cranking for aggregate many times, we don't need to do that anymore so you can make the test file shorter

program/rust/src/tests/test_upd_price_v2.rs
```diff
@@ -429,16 +421,86 @@ fn test_upd_price_v2() -> Result<(), Box<dyn std::error::Error>> {
...
+        );
+        assert_eq!(price_data.prev_price_cumulative.num_down_slots, 16);
         assert_eq!(
             price_data.price_cumulative.price,
-            3 * 42 + 81 * 3 - 100 * 3 + 42 * 60
+            42 + 81 * 4 - 100 * 4 + 60 * 41 + 55 + 14 * 49
+        ); // (42 + 81 * 4 - 100 * 4 + 60 * 41 + 55) is the price cumulative from the previous test and 14 * 49 (slot_gap) is the price cumulative from this test
+        assert_eq!(
+            price_data.price_cumulative.conf,
+            2 * 5 + 4 + 41 * 4 + 5 + 6 * 49
```
guibescos, line 501: Feels like you should test the behavior when there are two aggregations in one slot here

program/c/src/oracle/upd_aggregate.h
guibescos: let's leave it for a future PR

program/rust/src/tests/test_upd_price_v2.rs
cctdaniel: what do you mean specifically? this test is meant to simulate 2 aggregations in one slot

program/rust/src/tests/test_upd_price.rs
```diff
@@ -396,7 +425,111 @@ fn test_upd_price() {
...
+        &program_id,
+        &[
+            funding_account.clone(),
+            price_account.clone(),
+            clock_account.clone()
+        ],
+        &instruction_data
+    )
+    .is_ok());
+
```
guibescos, line 487: I'd put another round of asserts here

program/rust/src/tests/test_upd_price_v2.rs
guibescos: Yes but no asserts after the first aggregation

guibescos (approved):

program/rust/src/tests/test_upd_price_v2.rs
cctdaniel: will add asserts

ali-bahjati (approved):

PR #395, opened 2024-02-27
feat: add SetMaxLatency instruction

cctdaniel:


guibescos (dismissed):
Good, only nitpick is to update `test_sizes.rs` with `SetMaxLatencyArgs`, also no conventional commit

guibescos (approved):

PR #398, opened 2024-03-21
refactor: move upd_twap from upd_aggregate to rust upd_price

cctdaniel:
this is the first refactor in part of the same slot agg changes, intentionally keeping it small and limited to `upd_twap` so that it's easier to review

### DIFF
--- a/program/c/src/oracle/upd_aggregate.h
+++ b/program/c/src/oracle/upd_aggregate.h
@@ -134,14 +134,6 @@ static inline void upd_twap(
 // update aggregate price
 static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timestamp )
 {
-  // only re-compute aggregate in next slot
-  if ( slot <= ptr->agg_.pub_slot_ ) {
-    return false;
-  }
-
-  // get number of slots from last published valid price
-  int64_t agg_diff = ( int64_t )slot - ( int64_t )( ptr->last_slot_ );
-
   // Update the value of the previous price, if it had TRADING status.
   if ( ptr->agg_.status_ == PC_STATUS_TRADING ) {
     ptr->prev_slot_      = ptr->agg_.pub_slot_;
@@ -224,7 +216,6 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
   ptr->agg_.price_  = agg_price;
   ptr->agg_.conf_   = (uint64_t)agg_conf;
 
-  upd_twap( ptr, agg_diff );
   return true;
 }
 
--- a/program/rust/src/processor/upd_price.rs
+++ b/program/rust/src/processor/upd_price.rs
@@ -177,27 +177,32 @@ pub fn upd_price(
 
     // Try to update the aggregate
     #[allow(unused_variables)]
-    let mut aggregate_updated = false;
-
-    // NOTE: c_upd_aggregate must use a raw pointer to price
-    // data. Solana's `<account>.borrow_*` methods require exclusive
-    // access, i.e. no other borrow can exist for the account.
-    #[allow(unused_assignments)]
-    if clock.slot > latest_aggregate_price.pub_slot_ {
-        unsafe {
-            aggregate_updated = c_upd_aggregate(
+    let aggregate_updated = if clock.slot > latest_aggregate_price.pub_slot_ {
+        let updated = unsafe {
+            // NOTE: c_upd_aggregate must use a raw pointer to price
+            // data. Solana's `<account>.borrow_*` methods require exclusive
+            // access, i.e. no other borrow can exist for the account.
+            c_upd_aggregate(
                 price_account.try_borrow_mut_data()?.as_mut_ptr(),
                 clock.slot,
                 clock.unix_timestamp,
-            );
-            let agg_diff = {
-                (clock.slot as i64)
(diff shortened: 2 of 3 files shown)


program/c/src/oracle/upd_aggregate.h
```diff
@@ -134,14 +134,6 @@ static inline void upd_twap(
 // update aggregate price
 static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timestamp )
 {
-  // only re-compute aggregate in next slot
-  if ( slot <= ptr->agg_.pub_slot_ ) {
-    return false;
-  }
```
cctdaniel, line 140: this doesn't happen because of `if clock.slot > latest_aggregate_price.pub_slot_` check in rust `upd_price.rs`

program/rust/src/processor/upd_price.rs
```diff
@@ -179,6 +179,12 @@ pub fn upd_price(
     #[allow(unused_variables)]
     let mut aggregate_updated = false;
 
+    let agg_diff = {
+        (clock.slot as i64)
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.last_slot_
+                as i64
+    };
+
```
cctdaniel, line 187: taken from `upd_aggregate.h`

program/rust/src/tests/test_upd_aggregate.rs
```diff
@@ -100,8 +100,6 @@ fn test_upd_aggregate() {
 
         assert_eq!(price_data.agg_.price_, 100);
         assert_eq!(price_data.agg_.conf_, 10);
-        assert_eq!(price_data.twap_.val_, 100);
```
guibescos, line 103: Maybe better to keep the asserts and run `c_upd_twap` after `c_upd_aggregate` above

program/c/src/oracle/upd_aggregate.h
guibescos: Seems good to remove this since it will need to go away for same slot agg

program/rust/src/processor/upd_price.rs
```diff
@@ -190,6 +196,10 @@ pub fn upd_price(
                 clock.slot,
                 clock.unix_timestamp,
             );
+            c_upd_twap(
+                price_account.try_borrow_mut_data()?.as_mut_ptr(),
+                agg_diff, // Ensure slots_since_last_update is cast to i64, as expected by the function signature
```
guibescos, line 201: Comment seems weird

program/rust/src/processor/upd_price.rs
```diff
@@ -179,6 +179,12 @@ pub fn upd_price(
     #[allow(unused_variables)]
     let mut aggregate_updated = false;
 
+    let agg_diff = {
+        (clock.slot as i64)
+            - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?.last_slot_
```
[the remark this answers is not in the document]
guibescos: I recommend that instead you use `prev_slot_` for this computation.

guibescos (commented):
Would like to see other reviews

program/rust/src/tests/test_upd_aggregate.rs
cctdaniel: hmm we already have a `test_ema.rs` which runs `c_upd_twap` do we need it here again?

program/rust/src/processor/upd_price.rs
```diff
@@ -190,6 +190,12 @@ pub fn upd_price(
                 clock.slot,
                 clock.unix_timestamp,
             );
+            let agg_diff = {
+                (clock.slot as i64)
+                    - load_checked::<PriceAccount>(price_account, cmd_args.header.version)?
+                        .prev_slot_ as i64
```
ali-bahjati, line 196: so the `prev_slot` after `c_upd_aggregate` will remain the same as `last_slot`before calling `c_upd_aggregate`?

program/rust/src/processor/upd_price.rs
```diff
@@ -190,6 +190,12 @@ pub fn upd_price(
...
+            };
+            c_upd_twap(price_account.try_borrow_mut_data()?.as_mut_ptr(), agg_diff);
```
ali-bahjati, line 198: I think this function only gets called if the aggregate is updated.

program/rust/src/processor/upd_price.rs
cctdaniel: yup since both `prev_slot` and `last_slot` is only updated in `c_upd_aggregate` if `ptr->agg_.status_ == PC_STATUS_TRADING`

ali-bahjati (approved):

program/c/src/oracle/upd_aggregate.h
```diff
@@ -134,14 +134,6 @@ static inline void upd_twap(
...
-
-  // get number of slots from last published valid price
-  int64_t agg_diff = ( int64_t )slot - ( int64_t )( ptr->last_slot_ );
-
```
jayantk, line 144: something to watch out for going forward is the if statement below this comment which updates the previous price. There's a chance that we introduce a bug that updates this multiple times on a single slot, which would make prev_slot == agg_.pub_slot_

jayantk (approved):

PR #402, opened 2024-05-04
refactor: use ordered publisher list

ali-bahjati:
This change brings down the average cost of a price update without aggregation from 5k to 2k by moving from unordered list of publishers to an ordered list of publishers on doing binary search on lookup. It will fall back to linear search if the binary search fails.

Also add_publisher now tries to sort the list with 000... publisher as a migration path. The sort is implemented in a way to be efficient and small (in code size). There is an insertion sort upon each new publisher to make sure the list remains sorted. del_publisher won't change the order of the list so it will remain sorted.

The heavy code path for publishers now uses `sol_memcmp` (on sorting and lookup) to be more optimized (it costs 10 CU and apparently publisher equality is a few times more expensive). I checked sol_memcmp without change to sorted but the result was not as significant. Maybe later we can extract it out to a function, and use it everywhere.

### DIFF
--- a/program/rust/src/processor/add_publisher.rs
+++ b/program/rust/src/processor/add_publisher.rs
@@ -83,23 +83,21 @@ pub fn add_publisher(
         }
     }
 
-    let mut current_index: usize = try_convert(price_data.num_)?;
+    let current_index: usize = try_convert(price_data.num_)?;
     sol_memset(
         bytes_of_mut(&mut price_data.comp_[current_index]),
         0,
         size_of::<PriceComponent>(),
     );
     price_data.comp_[current_index].pub_ = cmd_args.publisher;
+    price_data.num_ += 1;
 
-    // Shift the element back to keep the publishers components sorted.
-    while current_index > 0
-        && price_data.comp_[current_index].pub_ < price_data.comp_[current_index - 1].pub_
+    // Sort the publishers in the list
     {
-        price_data.comp_.swap(current_index, current_index - 1);
-        current_index -= 1;
+        let num_comps = try_convert::<u32, usize>(price_data.num_)?;
+        sort_price_comps(&mut price_data.comp_, num_comps)?;
     }
 
-    price_data.num_ += 1;
     price_data.header.size = try_convert::<_, u32>(PriceAccount::INITIAL_SIZE)?;
     Ok(())
 }
@@ -107,7 +105,6 @@ pub fn add_publisher(
 /// A copy of rust slice/sort.rs heapsort implementation which is small and fast. We couldn't use
 /// the sort directly because it was only accessible behind a unstable feature flag at the time of
 /// writing this code.
-#[inline(always)]
 fn heapsort(v: &mut [(Pubkey, usize)]) {
     // This binary heap respects the invariant `parent >= child`.
     let sift_down = |v: &mut [(Pubkey, usize)], mut node: usize| {
@@ -155,16 +152,19 @@ fn heapsort(v: &mut [(Pubkey, usize)]) {
 ///
 /// num_publishers is the number of publishers in the list that should be sorted. It is explicitly
 /// passed to avoid callers mistake of passing the full slice which may contain uninitialized values.
-#[inline(always)]
 fn sort_price_comps(comps: &mut [PriceComponent], num_comps: usize) -> Result<(), ProgramError> {
     let comps = comps
         .get_mut(..num_comps)
         .ok_or(ProgramError::InvalidArgument)?;
 
+    // Publishers are likely sorted in ascending order but
+    // heapsorts creates a max-heap so we reverse the order
+    // of the keys to make the heapify step faster.
     let mut keys = comps
         .iter()
         .enumerate()
         .map(|(i, x)| (x.pub_, i))
+        .rev()
         .collect::<Vec<_>>();
 
     heapsort(&mut keys);
--- a/program/rust/src/processor/add_publisher.rs
+++ b/program/rust/src/processor/add_publisher.rs
@@ -24,7 +24,10 @@ use {
         account_info::AccountInfo,
         entrypoint::ProgramResult,
         program_error::ProgramError,
-        program_memory::sol_memset,
+        program_memory::{
+            sol_memcmp,
+            sol_memset,
+        },
         pubkey::Pubkey,
     },
     std::mem::size_of,
@@ -41,8 +44,7 @@ pub fn add_publisher(
     let cmd_args = load::<AddPublisherArgs>(instruction_data)?;
 
     pyth_assert(
-        instruction_data.len() == size_of::<AddPublisherArgs>()
-            && cmd_args.publisher != Pubkey::default(),
+        instruction_data.len() == size_of::<AddPublisherArgs>(),
         ProgramError::InvalidArgument,
     )?;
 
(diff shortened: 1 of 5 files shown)


program/rust/src/processor/add_publisher.rs
```diff
@@ -73,13 +73,21 @@ pub fn add_publisher(
...
-    let current_index: usize = try_convert(price_data.num_)?;
+    let mut current_index: usize = try_convert(price_data.num_)?;
     sol_memset(
         bytes_of_mut(&mut price_data.comp_[current_index]),
         0,
         size_of::<PriceComponent>(),
     );
     price_data.comp_[current_index].pub_ = cmd_args.publisher;
+
+    while current_index > 0
```
Reisen, line 84: Would suggest adding a comment, PR title is clear but good to explain logic like this in place as well.

program/rust/src/processor/upd_price.rs
```diff
@@ -127,27 +127,74 @@ pub fn upd_price(
...
+        // search might not work if the publisher list is not sorted; therefore
+        // we fall back to a linear search and during the search we try to bring
+        // the publishers list closer to a sorted state. (We do not sort them
+        // entirely to avoid exceeding the compute budget.)
+        let mut binary_search_result = None;
+
+        {
+            // Binary search to find the publisher key. Rust std binary search is not used because
+            // they guarantee valid outcome only if the array is sorted whereas we want to rely on
+            // a Equal match if it is a result on an unsorted array. Currently the rust
```
Reisen, line 151: Is this still true if even if you use the binary_search_by_key variants? My worry in the call was the the result can't be verified but you mentioned the quality should confirm. Just want to double check if the reason the removal those forms of binary_search* is for this reason? The quality match branch seems right.

ali-bahjati: Well the implementation is the same but the API doc doesn't guarantee that. That's why I want to just be on the safe side here by having our code.

ali-bahjati: To be specifc the rust interface doesn't guarantee that if it returns an Ok(index) the pub[index] == key and also it holds true based on their implementation, i don't want to rely on it (who knows, maybe they change it later and we don't notice it).

jayantk (commented):
I don't love the incremental sorting pass -- see inline comment -- but happy to discuss that if there's some particular reason you are doing it this way

program/rust/src/processor/upd_price.rs
```diff
@@ -127,27 +127,74 @@ pub fn upd_price(
...
+            // implementation behaves the same but we do not want to rely on api internals.
+            let mut left = 0;
```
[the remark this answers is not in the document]
ali-bahjati: I hesitated to do this because i don't know whether function call adds cu overhead. Let me do it with `inline` and if it's fine I'll add the test. That being said, in our test since all the publishers are publishing prices the binary search gets test for each publisher. I'll add a comment explaining it too.

program/rust/src/processor/upd_price.rs
```diff
@@ -127,27 +127,74 @@ pub fn upd_price(
...
         }
-        pyth_assert(
-            publisher_index < try_convert::<u32, usize>(price_data.num_)?,
-            OracleError::PermissionViolation.into(),
-        )?;
+
+        publisher_index = match binary_search_result {
+            Some(index) => index,
+            None => {
+                // One pass through the publishers list to make it closer
```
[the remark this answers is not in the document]
ali-bahjati: I think the main reason here is having zero overhead upgrade + least increase in cu for the unhappy paths. Also adding full sort to the publisher can cause an enormous increase in cu and can break xc-admin when we do the upgrade (again less maintenance).

program/rust/src/tests/test_upd_price_v2.rs
```diff
@@ -444,6 +444,90 @@ fn test_upd_price_v2() -> Result<(), Box<dyn std::error::Error>> {
...
+    {
+        let mut price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        price_data.num_ = 20;
+        // Store the publishers in reverse order
+        publishers
+            .iter()
+            .rev()
+            .enumerate()
+            .for_each(|(i, account)| {
+                println!("{:} {:?}", i, account.key);
```
guibescos, line 475: println!

ali-bahjati: oh no. my bad here.

jayantk (approved):

program/rust/src/processor/add_publisher.rs
```diff
@@ -73,14 +83,149 @@ pub fn add_publisher(
...
     );
     price_data.comp_[current_index].pub_ = cmd_args.publisher;
+
+    // Shift the element back to keep the publishers components sorted.
+    while current_index > 0
+        && price_data.comp_[current_index].pub_ < price_data.comp_[current_index - 1].pub_
+    {
+        price_data.comp_.swap(current_index, current_index - 1);
+        current_index -= 1;
+    }
```
jayantk, line 100: minor comment: imo, we should call `sort_price_comps` here as well, because the insertion code above isn't tested by the quickcheck stuff below. We don't add publishers a lot so we don't care about optimizing CU use.

jayantk (approved):

program/rust/src/processor/add_publisher.rs
jayantk: (feel free to ignore this. I'm only saying it because it feels like the maximally defensive implementation)

ali-bahjati: did that. removed the `inline` hint to make sure code size doesn't get larger (by sacrificing a bit of CU)

jayantk (approved):

Reisen (approved):

program/rust/src/tests/test_add_publisher.rs
```diff
@@ -163,4 +163,45 @@ fn test_add_publisher() {
...
+            permissions_account.clone(),
+        ],
+        instruction_data
+    )
+    .is_ok());
+
+    // Make sure that publishers get sorted after adding the default publisher
+    {
+        let price_data = load_checked::<PriceAccount>(&price_account, PC_VERSION).unwrap();
+        println!("{:?}", price_data.comp_.map(|x| x.pub_));
```
guibescos, line 202: print statement

program/rust/src/processor/upd_price.rs
```diff
@@ -289,3 +345,35 @@ struct MessageBufferAccounts<'a, 'b: 'a> {
...
+    };
+
+    /// Test the find_publisher_index method works with an unordered list of components.
+    #[quickcheck]
+    pub fn test_find_publisher_index_unordered_comp(comps: Vec<PriceComponent>) {
+        comps.iter().enumerate().for_each(|(idx, comp)| {
+            assert_eq!(find_publisher_index(&comps, &comp.pub_), Some(idx));
+        });
+
+        assert_eq!(find_publisher_index(&comps, &Pubkey::new_unique()), None);
```
guibescos, line 365: Is this technically wrong since `Pubkey::arbitrary()` can sometimes be equal to `Pubkey::new_unique()`

ali-bahjati: Well this one is game of probabilities, the chances should be nearly zero.

program/rust/src/processor/upd_price.rs
```diff
@@ -281,6 +281,62 @@ pub fn upd_price(
...
+            }
+        }
+    }
+
+    match binary_search_result {
+        Some(index) => Some(index),
+        None => {
+            let mut index = 0;
+            while index < comps.len() {
+                if sol_memcmp(comps[index].pub_.as_ref(), key.as_ref(), 32) == 0 {
```
guibescos, line 326: What's the reason for using memcmp here

guibescos (approved):

program/rust/src/processor/upd_price.rs
ali-bahjati: Described in the PR description. It saves more cost.

program/rust/src/processor/upd_price.rs
ali-bahjati: well it's not obvious actually. will fix it.

guibescos (approved):

PR #403, opened 2024-05-13
refactor: optimize some checks on upd_price

ali-bahjati:
- Remove rent check because it costs 500CU (+10% of upd_price without aggregate) and it really is not needed.
- Use abs() on checking confidence threshold because it's incredibly slow on negative prices (2000CU). We don't have negative prices but will guard us in the future.

### DIFF
--- a/program/rust/src/processor/upd_price.rs
+++ b/program/rust/src/processor/upd_price.rs
@@ -13,7 +13,7 @@ use {
         instruction::UpdPriceArgs,
         utils::{
             check_valid_funding_account,
-            check_valid_writable_account_without_rent_check,
+            check_valid_writable_account,
             get_status_for_conf_price_ratio,
             is_component_update,
             pyth_assert,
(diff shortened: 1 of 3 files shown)


jayantk (approved):

program/rust/src/processor/upd_price.rs
```diff
@@ -123,7 +123,7 @@ pub fn upd_price(
     }?;
 
     check_valid_funding_account(funding_account)?;
-    check_valid_writable_account(program_id, price_account)?;
+    check_valid_writable_account_without_rent_check(program_id, price_account)?;
```
guibescos, line 126: mmmmm what about removing the rent check everywhere? I think the check is not useful anywhere

guibescos: this is introducing more complexity

guibescos (approved):

PR #404, opened 2024-05-14
refactor: replace sorting with quick select

ali-bahjati:
This change replaces the deterministic optimized merge-sort with a highly optimized quick select that performs signficantly faster when publisher count is more than 20. The previous algorithm is very fast running on a normal CPU but doesn't work very well in BPF because all instructions (like load, move, ..) have the same cost as comparisons and are not optimized. Quick select on the other hand, minimized all the operations in the average case due to its minimal reads and writes.

In 32-publisher setup, quick select reduces the CU from 16.5k to 11.5k and in the 64-publisher setup 37k to 17.5k. The numbers are the worst cases running on randomized input.

Worst measured case happens when all elements are equal and is 13k and 24k for 32-publisher and 64-publisher setup respectively.

Unfortunately there is no way to systematically get the the compute usage out of program test. The `test_benchmark` file has a simple code that helps running benchmarks on various number of publishers.

Quick select algorithm has a quadratic worst case but it is very unlikely to happen and no publisher alone can manipulate the list of prices by themselves.


program/c/src/oracle/model/test_price_model.c
```diff
@@ -19,15 +19,14 @@ int test_price_model() {
...
 
-# define N 96
+# define N 192
 
   int64_t quote0 [N];
   int64_t quote  [N];
   int64_t val    [3];
   int64_t scratch[N];
 
   for( int iter=0; iter<10000000; iter++ ) {
```
ali-bahjati, line 29: This tests ensures that the code works with 10000000 random test cases.

program/c/src/oracle/model/price_model.c
```diff
@@ -1,99 +1,108 @@
...
+                if (a[n / 2] < a[n - 1]) pivot = a[n / 2];
+                else if (a[0] < a[n - 1]) pivot = a[n - 1];
+                else pivot = a[0];
+            } else {
+                if (a[n / 2] > a[n - 1]) pivot = a[n / 2];
+                else if (a[0] > a[n - 1]) pivot = a[n - 1];
+                else pivot = a[0];
+            }
+        }
+
```
Reisen, line 36: The median of 3 as pivot already reduces the number of comparisons for the rest of the algorithm, but I think this can be reduced even further by initially picking the middle point optimistically:

```
        uint64_t mid = n / 2;
        int64_t pivot = a[mid];
        if ((a[0] < pivot) != (a[0] < a[n - 1])) {
            pivot = a[0] < a[n - 1] ? a[n - 1] : a[0];
        }
```

Here, we choose the midpoint optimistically then the condition checks if `a[0]` lies between `mid`  and `a[n-i]`, if the check fails we already have the right choice, if not, then the ternary condition selects between `a[0]` and `a[n-1]` which is the same as the logic you already had.

Not writing this as a suggestion to change as I'm not sure if there's any benefit in this change in terms of reducing instruction count.

program/c/src/oracle/model/price_model.c
```diff
@@ -1,99 +1,108 @@
...
+                j--;
+                continue;
+            }
+
+            int64_t tmp = a[i];
+            a[i] = a[j];
+            a[j] = tmp;
+            i++;
+            j--;
+        }
```
Reisen, line 72: This is potentially overly nitpicking but you can reduce the instruction count slightly here:

```
while (i <= j) {
    if (a[i] < pivot) {
        i++;
    } else if (a[j] > pivot) {
        j--;
    } else {
        int64_t tmp = a[i];
        a[i] = a[j];
        a[j] = tmp;
        i++;
        j--;
    }
}
```

With the continue statements you force a loop condition when the loop could have exited, so the `continue` statements force a recheck of the while condition unnecessarily, I doubt it is going to matter that much compared to the improvements you've made but just mentioning, the if/else form might be more readable though.

program/c/src/oracle/model/price_model.c
Reisen: Never mind there's no benefit, your explicit checks are clearer. Ignore this.

program/c/src/oracle/model/price_model.c
Reisen: Same here no benefit, if the compiler can't optimize this I'd be shocked and even if it did it's only two instructions. Ignore.

ali-bahjati: I checked it, the result was the same. Probably compiler does exactly what you said.

PR #405, opened 2024-05-15
refactor: replace merge-sort with heapsort

ali-bahjati:
This change replaces the merge-sort with a heapsort that uses much less CU than merge-sort.

The previous algorithm is very fast running on a normal CPU but doesn't work very well in BPF because all instructions (like load, move, ..) have the same cost and there is no cache and optimizations for memory-alignment have no real benefit.

The major benefits of heapsort are being non-recursive that reduces the high stackframe overhead in BPF and being inplace which minimizes number of copies.

Unfortunately there is no way to systematically get the the compute usage out of program test. The `test_benchmark` file has a simple code that helps running benchmarks on various number of publishers.

In 32-publisher setup, heapsort reduces the CU from 16.5k to 12k and in the 64-publisher setup 37k to 20.5k. The numbers are the worst cases running on randomized input. The result of running in a highly similar input (two distinct prices, and two distinct confidences) is 14.7k and 25k respectively, which is still better and is very unlikely in practice.


program/c/src/oracle/sort/tmpl/sort_stable.c
```diff
@@ -1,195 +0,0 @@
-/* Usage:
```
jayantk, line 1: you deleted this one because the sort is no longer stable (?)

(I don't think it matters that the sort is stable so that's fine)

program/c/src/oracle/model/price_model.h
```diff
@@ -8,91 +8,12 @@
...
-
-   The cost of this function is a fast and low variance (but not
-   completely data oblivious) O(cnt lg cnt) in the best / average /
-   worst cases.  This function uses no heap / dynamic memory allocation.
-   It is thread safe provided it passed non-conflicting quote, output
-   and scratch arrays.  It has a bounded call depth ~lg cnt <= ~64 (this
-   could reduce to O(1) by using a non-recursive sort/select
-   implementation under the hood if desired). */
-
-int64_t *                              /* Returns pointer to sorted quotes (either quote or ALIGN_UP(scratch,int64_t)) */
```
jayantk, line 74: suggest documenting that this mutates the contents of the `quote` array. I think that's the same behavior as the previous sort, but may not be obvious to readers in the future.

jayantk (approved):
very nicely done

program/c/src/oracle/sort/tmpl/sort_stable.c
ali-bahjati: Yeah. even being sorted is not important.

program/c/src/oracle/upd_aggregate.h
```diff
@@ -188,8 +188,7 @@ static inline bool upd_aggregate( pc_price_t *ptr, uint64_t slot, int64_t timest
     // note: numv>0 and nprcs = 3*numv at this point
     int64_t agg_p25;
     int64_t agg_p75;
-    int64_t scratch[ PC_NUM_COMP * 3 ]; // ~0.75KiB for current PC_NUM_COMP (FIXME: DOUBLE CHECK THIS FITS INTO STACK FRAME LIMIT)
```
guibescos, line 191: Great that it's in place now

program/c/src/oracle/sort/test_sort_stable.c
```diff
@@ -1,73 +0,0 @@
...
-        if( j<0 || j>=n || z[i]!=w[j] ) { printf( "FAIL (corrupt)\n" ); return 1; }
-        w[j] = -1; /* Mark that this entry has already been confirmed */
-      }
-      for( int i=0; i<n; i++ ) if( w[i]!=-1 ) { printf( "FAIL (perm)\n" ); return 1; }
-
-      /* Make sure that z is in order and stable */
-      for( int i=1; i<n; i++ )
-        if( z[i]<=z[i-1] ) { printf( "FAIL (%s, b=%lx)\n", BEFORE( z[i], z[i-1] ) ? "order" : "stable", b ); return 1; }
-    }
-  }
```
guibescos, line 41: This test was very nice, can we do something similar.
(Testing the price_model on all arrays with 0 or 1 as values).

ali-bahjati: Added.

program/c/src/oracle/model/test_price_model.c
```diff
@@ -19,13 +19,38 @@ int test_price_model() {
...
   int64_t quote  [N];
   int64_t val    [3];
-  int64_t scratch[N];
 
+  /* Brute force validate small sizes via the 0-1 principle. */
+  for( int cnt=0; cnt<=24; cnt++ ) {
+    for( long mask=0L; mask<(1L<<cnt); mask++ ) {
+      for( int i=0; i<cnt; i++ ) quote[i] = (int64_t) ((mask>>i) & 1L);
+
+      memcpy( quote, quote0, sizeof(int64_t)*(size_t)cnt );
```
Reisen, line 33: Is `memcpy` here different from the standard library in argument order? It looks like this overwrites the binary mask you just created with what is likely garbage memory. Shouldn't the order be reversed? If I understand the test right this would still pass even if this is backwards but it means the test isn't actually using the 0-1 principle.

Reisen (approved):

program/c/src/oracle/model/test_price_model.c
ali-bahjati: Oh yeah, my bad here. Thanks for catching it.

guibescos (approved):
tyvm

PR #407, opened 2024-06-07
feat(oracle): WIP add scores account

keyvankhademi:


program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,158 @@
...
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PublisherScoresAccount {
+    pub header:                AccountHeader,
+    pub num_publishers:        usize,
+    pub num_symbols:           usize,
+    // a constant used to normalize the scores
+    pub z:                     u32,
+    pub publisher_permissions: [[bool; PC_MAX_SYMBOLS as usize]; PC_MAX_PUBLISHERS as usize],
```
ali-bahjati, line 30: `[bool; usize]` is like C++ then it'll take usize bytes, i recommend using u64 (or word size) and accessing it (or external library if program size allows it)

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub scores:                [f64; PC_MAX_PUBLISHERS as usize],
```
ali-bahjati, line 31: I am not sure that in BPF we have floating point operations. iirc people used fixed point decimal operations.

program/c/src/oracle/oracle.h
```diff
@@ -21,6 +21,8 @@ extern "C" {
 #define PC_PUBKEY_SIZE       32
 #define PC_PUBKEY_SIZE_64   (PC_PUBKEY_SIZE/sizeof(uint64_t))
 #define PC_MAP_TABLE_SIZE   640
+#define PC_MAX_PUBLISHERS    256
+#define PC_MAX_SYMBOLS       1024
```
ali-bahjati, line 25: what we do if we exceed this numbers?

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,158 @@
...
+                .sum();
+        }
+        Ok(())
+    }
+}
+
+impl PythAccount for PublisherScoresAccount {
+    const ACCOUNT_TYPE: u32 = PC_ACCTYPE_SCORE;
+    // Calculate the initial size of the account
+    const INITIAL_SIZE: u32 = 1000;
```
ali-bahjati, line 150: this one doesn't seem right?

program/c/src/oracle/oracle.h
guibescos: That sounds like a Pythnet v2 problem anyway

program/rust/src/accounts/score.rs
guibescos: I also lean towards fixed point

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Unsafe impl because product_list is of size 640 and there's no derived trait for this size
+unsafe impl Pod for PublisherScoresAccount {
```
guibescos, line 177: The problem with Pod is the z as u32 adds some padding, so it can't be Pod anymore.
If you make z an u64 then you can use the derive macro

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub num_publishers: usize,
+    pub num_symbols:    usize,
+    // a constant used to normalize the scores
+    pub z:              u32,
+
+    // array[x][y] is a u64 whose bits represent if publisher x publishes symbols 64*y to 64*(y+1) - 1
+    pub publisher_permissions: [[u64; PC_MAX_SYMBOLS_64 as usize]; PC_MAX_PUBLISHERS as usize],
+    pub scores:                [f64; PC_MAX_PUBLISHERS as usize],
+    pub publishers:            [Pubkey; PC_MAX_PUBLISHERS as usize],
+    pub symbols:               [Pubkey; PC_MAX_SYMBOLS as usize],
```
guibescos, line 36: We actually already have something similar to this (a list of all symbols) in the mapping account but the product accounts (and not the price accounts) are listed there

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn get_publisher_permission(&self, x: usize, y: usize) -> bool {
+        (self.publisher_permissions[x][y / 64] >> (y % 64)) & 1 == 1
+    }
+
+    fn set_publisher_permission(&mut self, x: usize, y: usize, value: bool) {
+        if value {
+            self.publisher_permissions[x][y / 64] |= 1 << (y % 64);
+        } else {
+            self.publisher_permissions[x][y / 64] &= !(1 << (y % 64));
+        }
```
guibescos, line 49: `get_publisher_permission` and `set_publisher_permission` could almost be in their own struct PermissionTable, that we can proptest

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub z:              u32,
+
+    // array[x][y] is a u64 whose bits represent if publisher x publishes symbols 64*y to 64*(y+1) - 1
+    pub publisher_permissions: [[u64; PC_MAX_SYMBOLS_64 as usize]; PC_MAX_PUBLISHERS as usize],
+    pub scores:                [f64; PC_MAX_PUBLISHERS as usize],
+    pub publishers:            [Pubkey; PC_MAX_PUBLISHERS as usize],
+    pub symbols:               [Pubkey; PC_MAX_SYMBOLS as usize],
+}
+
+impl PublisherScoresAccount {
```
guibescos, line 39: This struct is begging for a prop test like in this https://github.com/pyth-network/pyth-client/blob/main/program/rust/src/tests/test_twap.rs#L50

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub fn calculate_scores(&mut self) -> Result<(), ProgramError> {
+        let symbol_scores: Vec<u32> = self
+            .symbols
+            .iter()
+            .enumerate()
+            .map(|(j, _)| {
+                let score = self
+                    .publisher_permissions
+                    .iter()
+                    .fold(0, |score, permissions| score + permissions[j] as u32);
```
guibescos, line 152: I think this doesn't work well with your logic for storing the booleans in a u64s

program/rust/src/processor/add_price.rs
```diff
@@ -48,10 +50,12 @@ pub fn add_price(
...
 
 
-    let (funding_account, product_account, price_account, permissions_account) = match accounts {
-        [x, y, z, p] => Ok((x, y, z, p)),
-        _ => Err(OracleError::InvalidNumberOfAccounts),
-    }?;
+    let (funding_account, product_account, price_account, permissions_account, scores_account) =
+        match accounts {
+            [x, y, z, p] => Ok((x, y, z, p, None)),
+            [x, y, z, p, s] => Ok((x, y, z, p, Some(s))),
```
guibescos, line 56: This part is tricky because:
a. Does it go out of sync if we don't pass the account
b. How does the program know s is _the_ score account, or can there be many score accounts?

guibescos: Maybe it's ok since this is a permissioned operation on a second thought

program/rust/src/tests/test_add_price.rs
```diff
@@ -62,6 +63,9 @@ fn test_add_price() {
     let mut permissions_setup = AccountSetup::new_permission(&program_id);
     let permissions_account = permissions_setup.as_account_info();
 
+    let mut scores_setup = AccountSetup::new::<PublisherScoresAccount>(&program_id);
+    let scores_account = scores_setup.as_account_info();
```
guibescos, line 67: There will need to be a way to initialize this account in the program:
1. If the account is the PDA (like the permssions account) you're limited to 10kb at creation but you can resize it later (increase it by 10kb) at a time
2. If it's a regular account the limit is 10MB. The problem is the address will be at a random address (instead of deterministically derived from a seed).

program/rust/src/accounts/score.rs
```diff
@@ -0,0 +1,188 @@
...
+    std::cmp::max,
+};
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+/*
+ * This account is part of Community Integrity Pool (CIP) project.
+ * It is used to store the caps of the publishers which will be sent
+ * to the `integrity_pool` program on mainnet to calculate the rewards.
+ */
```
ali-bahjati, line 29: we kind of use `///` in rust everywhere.

program/rust/src/accounts/score.rs
keyvankhademi: Yeah I guessed it might be the same. Fixed. Thank you.

program/rust/src/accounts/score.rs
keyvankhademi: Changed it to u64 👍.

program/rust/src/accounts/score.rs
keyvankhademi: it was a placeholder at the time. Fixed.

program/rust/src/accounts/score.rs
keyvankhademi: I changed them to u64, it simplified a lot of castings as well.

program/rust/src/accounts/score.rs
keyvankhademi: I will try this at some point to see how much it will change the program size, my current guess is that it will increase it.

program/rust/src/accounts/score.rs
keyvankhademi: Fixed.

program/rust/src/processor/add_price.rs
keyvankhademi: It is also just for backward compatibility and will remove the first clause when proposals have been migrated.

program/rust/src/tests/test_add_price.rs
keyvankhademi: we probably will go with number 2. We can also hardcode the account to avoid duplication.

program/rust/src/accounts/score.rs
keyvankhademi: Fixed.

program/c/src/oracle/oracle.h
keyvankhademi: The program doesn't allow exceeding these numbers, we will probably face a lot of issues if we decide to increase these numbers. We are hoping to go to pythnet V2 before we need to increase these numbers. From the trasactions limit that we currently have, i don't think this will an issue.
But if you think there is a chance that we exceed these number before pythnet v2, we better address it now.

program/rust/src/processor/add_publisher.rs
```diff
@@ -93,139 +102,26 @@ pub fn add_publisher(
     price_data.num_ += 1;
 
     // Sort the publishers in the list
```
ali-bahjati, line 104: Update this comment

PR #410, opened 2024-07-09
Support aggregation in validator

Riateche:
- Change dependency versions to make the crate compatible with pythnet validator
- Add flags to price account to track transition from v1 to v2
- Implement transition between v1 and v2 by calling add_publisher with a special pubkey
- Skip aggregation and message buffer update in the oracle program if v2 aggregation is enabled
- Clear message buffers after transition to v2
- Expose aggregation as a library function to be used in the validator

### DIFF
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -123,7 +123,6 @@ mod price_pythnet {
         }
         /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
         pub fn update_price_cumulative(&mut self) {
-            debug_assert!(self.agg_.status_ == PC_STATUS_TRADING);
             if self.agg_.status_ == PC_STATUS_TRADING {
                 self.price_cumulative.update(
                     self.agg_.price_,
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -98,7 +98,10 @@ mod price_pythnet {
         #[derive(Copy, Clone, Pod, Zeroable)]
         pub struct PriceAccountFlags: u8 {
             /// If set, the program doesn't do accumulation, but validator does.
-            const ACCUMULATOR_V2 = 1;
+            const ACCUMULATOR_V2 = 0b1;
+            /// If unset, the program will remove old messages from its message buffer account
+            /// and set this flag.
+            const MESSAGE_BUFFER_CLEARED = 0b10;
         }
     }
 
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -33,6 +33,7 @@ mod price_pythnet {
             },
             error::OracleError,
         },
+        bitflags::bitflags,
     };
 
     /// Pythnet-only extended price account format. This extension is
@@ -65,8 +66,9 @@ mod price_pythnet {
         pub message_sent_:      u8,
         /// Configurable max latency in slots between send and receive
         pub max_latency_:       u8,
+        /// Various flags
+        pub flags:              PriceAccountFlags,
         /// Unused placeholder for alignment
-        pub unused_2_:          i8,
         pub unused_3_:          i32,
         /// Corresponding product account
         pub product_account:    Pubkey,
@@ -89,6 +91,15 @@ mod price_pythnet {
         pub comp_:              [PriceComponent; PC_NUM_COMP_PYTHNET as usize],
         /// Cumulative sums of aggregative price and confidence used to compute arithmetic moving averages
         pub price_cumulative:   PriceCumulative,
+    }
+
+    bitflags! {
+        #[repr(C)]
+        #[derive(Copy, Clone, Pod, Zeroable)]
+        pub struct PriceAccountFlags: u8 {
+            /// If set, the program doesn't do accumulation, but validator does.
+            const ACCUMULATOR_V2 = 1;
+        }
     }
 
     impl PriceAccountPythnet {
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -25,13 +25,10 @@ mod price_pythnet {
 
     use {
         super::*,
-        crate::{
-            c_oracle_header::{
-                PC_MAX_SEND_LATENCY,
-                PC_NUM_COMP_PYTHNET,
-                PC_STATUS_TRADING,
-            },
-            error::OracleError,
+        crate::c_oracle_header::{
+            PC_MAX_SEND_LATENCY,
+            PC_NUM_COMP_PYTHNET,
+            PC_STATUS_TRADING,
         },
         bitflags::bitflags,
     };
@@ -125,7 +122,8 @@ mod price_pythnet {
             }
         }
         /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
-        pub fn update_price_cumulative(&mut self) -> Result<(), OracleError> {
+        pub fn update_price_cumulative(&mut self) {
+            debug_assert!(self.agg_.status_ == PC_STATUS_TRADING);
             if self.agg_.status_ == PC_STATUS_TRADING {
                 self.price_cumulative.update(
                     self.agg_.price_,
@@ -133,9 +131,6 @@ mod price_pythnet {
                     self.agg_.pub_slot_.saturating_sub(self.prev_slot_),
                     self.max_latency_,
                 ); // pub_slot should always be >= prev_slot, but we protect ourselves against underflow just in case
-                Ok(())
-            } else {
-                Err(OracleError::NeedsSuccesfulAggregation)
             }
         }
 
--- a/program/rust/src/processor/add_publisher.rs
+++ b/program/rust/src/processor/add_publisher.rs
@@ -82,7 +82,9 @@ pub fn add_publisher(
         price_data.flags.insert(PriceAccountFlags::ACCUMULATOR_V2);
         return Ok(());
     } else if cmd_args.publisher == Pubkey::from(DISABLE_ACCUMULATOR_V2) {
-        price_data.flags.remove(PriceAccountFlags::ACCUMULATOR_V2);
+        price_data
+            .flags
+            .remove(PriceAccountFlags::ACCUMULATOR_V2 | PriceAccountFlags::MESSAGE_BUFFER_CLEARED);
         return Ok(());
     }
 
--- a/program/rust/src/processor/add_publisher.rs
+++ b/program/rust/src/processor/add_publisher.rs
@@ -2,6 +2,7 @@ use {
     crate::{
         accounts::{
             PriceAccount,
+            PriceAccountFlags,
             PriceComponent,
             PythAccount,
         },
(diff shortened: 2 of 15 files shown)


program/rust/src/validator.rs
```diff
@@ -0,0 +1,131 @@
...
+    AlreadyAggregated,
+}
+
+#[derive(Debug)]
+pub struct AggregationOutcome {
+    pub messages: [Vec<u8>; 2],
+    pub commit:   bool,
+}
+
+pub fn aggregate_price(
```
jayantk, line 111: Please document this method. Few questions that the reader will have are:
1) under what conditions is `price_account_data` updated?
2) what does the `commit` flag on the return type mean?

program/rust/src/validator.rs
```diff
@@ -0,0 +1,131 @@
...
+    unsafe {
+        c_upd_twap(price_account as *mut PriceAccount as *mut u8, agg_diff);
+    }
+
+    // We want to send a message every time the aggregate price updates. However, during the migration,
+    // not every publisher will necessarily provide the accumulator accounts. The message_sent_ flag
+    // ensures that after every aggregate update, the next publisher who provides the accumulator accounts
+    // will send the message.
+    price_account.message_sent_ = 0;
+    if price_account.update_price_cumulative().is_err() {
```
jayantk, line 87: This `if` statement seems like it's a change in the logic relative to the prior implementation, and I'm not sure this is right. There are cases where aggregation can be successful and the status is unknown (e.g., not enough publisher prices are < 25 slots old).

I'm not sure what it means to "commit" here, but this seems suspicious to me.

Riateche: I believe this code reproduces the original behavior. As seen [here](https://github.com/pyth-network/pyth-client/blob/main/program/rust/src/processor/upd_price.rs#L194), `upd_price` instruction handler calls `update_price_cumulative()` and passes through the error, so the transaction will revert in case `update_price_cumulative()` fails.

program/rust/src/validator.rs
Riateche: Added some docs.

program/rust/src/processor/add_publisher.rs
```diff
@@ -71,6 +78,14 @@ pub fn add_publisher(
         let num_comps = try_convert::<u32, usize>(price_data.num_)?;
         sort_price_comps(&mut price_data.comp_, num_comps)?;
         return Ok(());
+    } else if cmd_args.publisher == Pubkey::from(ENABLE_ACCUMULATOR_V2) {
```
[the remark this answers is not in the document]
Riateche: Done.

program/rust/src/validator.rs
jayantk: ok I missed the `?` in the `upd_price` implementation above, so you're right about this.

However, I think the if statement on line 71 above is wrong then. `c_upd_aggregate` mutates the price account content regardless of whether it returns `true` or `false`. (If it returns `false`, it sets the status flag in the account to `UNKNOWN`. If `true`, the status is `TRADING`. Either way, that change is persisted.) This update cumulative call in the original implementation is gated by the `updated` flag specifically because that guarantees that the status is `TRADING` (and thus the call won't return an error)

jayantk: (as a meta-point, I believe this implementation of the function guarantees that the status flag in the price account can never be set to UNKNOWN. That's a behavior change relative to the current aggregation, so something is wrong somewhere.)

jayantk: another thought here is maybe could use the existing tests of aggregation to validate that the new logic is the same as the old logic? A janky approach would be to copy/paste `test_upd_price` and then in each case there where it loads the price account to assert its contents, call this function on the PriceAccount before running the assertions.

An approach like that would give you lots of confidence in the implementation.

ali-bahjati (commented):
I think as Jayant said, would be nice to test the validator update code path too.

program/rust/src/validator.rs
Riateche: > However, I think the if statement on line 71 above is wrong then. c_upd_aggregate mutates the price account content regardless of whether it returns true or false. (If it returns false, it sets the status flag in the account to UNKNOWN. If true, the status is TRADING. Either way, that change is persisted.)

Yeah, that's a good catch. I've fixed the code to not revert when `c_upd_aggregate` returns false. We'll work on adding more tests as you suggested.

> I think we should fix the behaviour on the current aggregation too.

Do you mean we should be ignoring the error returned by `update_price_cumulative()` in the contract as well? Or maybe we should change `update_price_cumulative()` to never return an error instead?

program/rust/src/validator.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+    let commit = update_aggregate(slot, timestamp, price_account);
+    let messages = [
+        price_account
+            .as_price_feed_message(price_account_pubkey)
+            .to_bytes(),
+        price_account
+            .as_twap_message(price_account_pubkey)
+            .to_bytes(),
+    ];
```
ali-bahjati, line 139: I think we need to add a check to not add it if the message buffer is not cleared. It can result it not sending the same account state temporarily until a new tx to update price comes but i think it's better than having duplicate messages.

program/rust/src/validator.rs
ali-bahjati: I don't have any strong opinion. I think either is fine. I rather not returning an error because it's kind of expected behaviour and cascading it can go unnoticed (like it did now)

Riateche: Changed `update_price_cumulative()` to not return an error.

program/rust/src/validator.rs
```diff
@@ -0,0 +1,140 @@
...
+
+#[derive(Debug)]
+pub struct AggregationOutcome {
+    /// Messages that should be included in the merkle tree
+    /// (a price feed message and a TWAP message).
+    pub messages: [Vec<u8>; 2],
+    /// A flag indicating that the changes to `price_account_data`
+    /// should be saved to the price account.
+    pub commit:   bool,
+}
```
ali-bahjati, line 112: We don't need it anymore now.

program/rust/src/accounts/price.rs
```diff
@@ -111,17 +122,15 @@ mod price_pythnet {
             }
         }
         /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
-        pub fn update_price_cumulative(&mut self) -> Result<(), OracleError> {
+        pub fn update_price_cumulative(&mut self) {
+            debug_assert!(self.agg_.status_ == PC_STATUS_TRADING);
```
ali-bahjati, line 126: hmmm why would we need the debug assert here :?

Riateche: Our assumption is that this check should always succeed when we call `update_price_cumulative()`, no?

Riateche: Removed debug assert.

program/rust/src/processor/add_publisher.rs
```diff
@@ -62,14 +70,19 @@ pub fn add_publisher(
...
 
-
     let mut price_data = load_checked::<PriceAccount>(price_account, cmd_args.header.version)?;
 
-    // Use the call with the default pubkey (000..) as a trigger to sort the publishers as a
-    // migration step from unsorted list to sorted list.
-    if cmd_args.publisher == Pubkey::default() {
-        let num_comps = try_convert::<u32, usize>(price_data.num_)?;
-        sort_price_comps(&mut price_data.comp_, num_comps)?;
+    if cmd_args.publisher == Pubkey::from(ENABLE_ACCUMULATOR_V2) {
```
guibescos, line 75: Quite frankly I like a new instruction better

program/rust/src/accounts/price.rs
```diff
@@ -111,17 +122,14 @@ mod price_pythnet {
             }
         }
         /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
-        pub fn update_price_cumulative(&mut self) -> Result<(), OracleError> {
+        pub fn update_price_cumulative(&mut self) {
```
guibescos, line 125: Why this change?

program/rust/src/validator.rs
```diff
@@ -0,0 +1,130 @@
...
+        return Err(AggregationError::V1AggregationMode);
+    }
+
+    Ok(data)
+}
+
+fn update_aggregate(slot: u64, timestamp: i64, price_account: &mut PriceAccount) {
+    // NOTE: c_upd_aggregate must use a raw pointer to price data. We already
+    // have the exclusive mut reference so we can simply cast before calling
+    // the function.
```
guibescos, line 67: imo you need `if clock.slot > latest_aggregate_price.pub_slot_` here to avoid double aggregation on the slot were the flag gets turned on

guibescos:
I think there might be an edgecase where the price stops aggregating (both v1 and v2).
In a given slot:
- `upd_price` : aggregation is unsuccessful therefore `message_sent_ == 1`
- flags gets flipped
- aggregate v2 fails with V1AggregationMode since `PriceAccountFlags::MESSAGE_BUFFER_CLEARED` is false

Next slot:
- `upd_price` no longer triggers an aggregation because `PriceAccountFlags::ACCUMULATOR_V2` is true
- aggregate v2 also fails since `PriceAccountFlags::MESSAGE_BUFFER_CLEARED` is false

And so on...

Riateche:
> I think there might be an edgecase where the price stops aggregating (both v1 and v2).

Good catch! I've changed the code so that the `MESSAGE_BUFFER_CLEARED` is set regardless of the `message_sent_` flag. I think this should fix the issue.

program/rust/src/tests/test_upd_price_with_validator.rs
```diff
@@ -0,0 +1,380 @@
...
+        assert_eq!(price_data.agg_.status_, PC_STATUS_UNKNOWN);
+    }
+
+    // We aggregate the price at the end of each slot now.
+    validator::aggregate_price(1, 101, price_account.key, *price_account.data.borrow_mut())
+        .unwrap();
+    update_clock_slot(&mut clock_account, 2);
+
+    validator::aggregate_price(2, 102, price_account.key, *price_account.data.borrow_mut())
+        .unwrap();
```
ali-bahjati, line 128: nit: can we once verify the output of this as well?

ali-bahjati (approved):
LGTM!

program/rust/src/processor/upd_price.rs
```diff
@@ -210,76 +210,74 @@ pub fn upd_price(
...
-                        .flags
-                        .insert(PriceAccountFlags::MESSAGE_BUFFER_CLEARED);
-                }
+                        .as_price_feed_message(price_account.key)
+                        .to_bytes(),
+                    price_data.as_twap_message(price_account.key).to_bytes(),
+                ]
+            };
+
+            // Append a TWAP message if available
```
jayantk, line 259: this comment seems like it doesn't apply to anything

jayantk (approved):

program/rust/src/tests/test_upd_price_with_validator.rs
Riateche: Added some output checks.

program/rust/src/processor/upd_price.rs
Riateche: Looks like it's no longer relevant. Removed.

ali-bahjati (dismissed):

ali-bahjati (approved):

PR #412, opened 2024-07-16
feat: add publisher caps to the validator

guibescos:
I implements the publisher cap as: 

image

where M and Z are parameters, n_s is the number of publishers of symbol s

### DIFF
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -123,7 +123,6 @@ mod price_pythnet {
         }
         /// This function gets triggered when there's a succesful aggregation and updates the cumulative sums
         pub fn update_price_cumulative(&mut self) {
-            debug_assert!(self.agg_.status_ == PC_STATUS_TRADING);
             if self.agg_.status_ == PC_STATUS_TRADING {
                 self.price_cumulative.update(
                     self.agg_.price_,
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -98,7 +98,10 @@ mod price_pythnet {
         #[derive(Copy, Clone, Pod, Zeroable)]
         pub struct PriceAccountFlags: u8 {
             /// If set, the program doesn't do accumulation, but validator does.
-            const ACCUMULATOR_V2 = 1;
+            const ACCUMULATOR_V2 = 0b1;
+            /// If unset, the program will remove old messages from its message buffer account
+            /// and set this flag.
+            const MESSAGE_BUFFER_CLEARED = 0b10;
         }
     }
 
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -13,7 +13,7 @@ use {
     },
     pythnet_sdk::messages::{
         PriceFeedMessage,
-        PublisherCapsMessage,
+        PublisherStakeCapsMessage,
         TwapMessage,
     },
     solana_program::pubkey::Pubkey,
@@ -342,7 +342,7 @@ impl PythOracleSerialize for TwapMessage {
     }
 }
 
-impl PythOracleSerialize for PublisherCapsMessage {
+impl PythOracleSerialize for PublisherStakeCapsMessage {
     fn to_bytes(self) -> Vec<u8> {
         const DISCRIMINATOR: u8 = 2;
         let mut result: Vec<u8> = Vec::new();
--- a/program/rust/src/accounts/price.rs
+++ b/program/rust/src/accounts/price.rs
@@ -345,9 +345,7 @@ impl PythOracleSerialize for TwapMessage {
 impl PythOracleSerialize for PublisherStakeCapsMessage {
     fn to_bytes(self) -> Vec<u8> {
         const DISCRIMINATOR: u8 = 2;
-        let mut result: Vec<u8> = Vec::new();
-
-        result.extend_from_slice(&[DISCRIMINATOR]);
+        let mut result = vec![DISCRIMINATOR];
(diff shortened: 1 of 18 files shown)


program/rust/src/validator.rs
```diff
@@ -128,3 +141,45 @@ pub fn aggregate_price(
...
 }
+
+fn checked_load_price_account(price_account_info: &[u8]) -> Option<&PriceAccount> {
+    check_price_account_header(price_account_info).ok()?;
+    Some(bytemuck::from_bytes::<PriceAccount>(
+        &price_account_info[0..size_of::<PriceAccount>()],
+    ))
+}
+
+pub fn compute_publisher_stake_caps(
```
ali-bahjati, line 152: Please document the arguments of this function.

program/rust/src/validator.rs
```diff
@@ -128,3 +141,45 @@ pub fn aggregate_price(
...
+    account_datas: Vec<&[u8]>,
```
Riateche, line 153: Could we maybe use `impl IntoIterator<Item = &[u8]>` instead to avoid an extra Vec?

program/rust/src/validator.rs
```diff
@@ -128,3 +141,45 @@ pub fn aggregate_price(
...
+    timestamp: i64,
+    m: u64,
+    z: u64,
```
Riateche, line 156: Can we rename `m` and `z` to something meaningful? `z` could be `max_publishers` or `num_publishers_cap`. Not sure about `m`.

Riateche (changes requested):

program/rust/src/validator.rs
guibescos: done

program/rust/src/validator.rs
guibescos: I like these names for simplicity since they're are also the name of the variables in the math...
There's also no good way of describing z.
I added a comment explaining them.

program/rust/src/validator.rs
guibescos: Done

program/rust/src/tests/test_message.rs
```diff
@@ -38,3 +43,25 @@ fn test_twap_message_roundtrip(input: TwapMessage) -> bool {
...
     }
 }
+
+
+fn prop_publisher_caps_message_roundtrip(input: PublisherStakeCapsMessage) -> bool {
+    let reconstructed = from_slice::<BigEndian, Message>(&input.clone().to_bytes()).unwrap();
+
+    println!("Failed test case:");
+    println!("{:?}", input);
+    println!("{:?}", reconstructed);
```
keyvankhademi, line 53: I think we can remove these.

keyvankhademi (dismissed):

program/rust/src/validator.rs
```diff
@@ -128,3 +141,53 @@ pub fn aggregate_price(
...
+) -> Vec<u8> {
+    let mut publisher_caps: BTreeMap<Pubkey, u64> = BTreeMap::new(); // BTreeMap to ensure it will be sorted by publisher
+    for account in account_datas {
+        if let Some(price_account) = checked_load_price_account(account) {
+            let cap: u64 = m
+                .checked_div(max(u64::from(price_account.num_), z))
+                .unwrap_or(0);
+            for i in 0..(price_account.num_ as usize) {
+                publisher_caps
+                    .entry(price_account.comp_[i].pub_)
```
Riateche, line 174: This is pretty unlikely to matter but it's safer to use checked indexing here of check `num_` beforehand to avoid crashing the validator if `num_` somehow ends up incorrect.

Riateche (dismissed):

program/rust/src/validator.rs
```diff
@@ -128,3 +141,53 @@ pub fn aggregate_price(
...
+    account_datas: impl IntoIterator<Item = &'a [u8]>,
+    timestamp: i64,
+    m: u64,
+    z: u64,
+) -> Vec<u8> {
+    let mut publisher_caps: BTreeMap<Pubkey, u64> = BTreeMap::new(); // BTreeMap to ensure it will be sorted by publisher
+    for account in account_datas {
+        if let Some(price_account) = checked_load_price_account(account) {
+            let cap: u64 = m
+                .checked_div(max(u64::from(price_account.num_), z))
```
ali-bahjati, line 170: side-note: it's good to update our formula in the doc to mention we take the floor of the division when we sum it up.

program/rust/src/accounts/price.rs
```diff
@@ -340,3 +344,23 @@ impl PythOracleSerialize for TwapMessage {
         bytes.to_vec()
     }
 }
+
+impl PythOracleSerialize for PublisherStakeCapsMessage {
+    fn to_bytes(self) -> Vec<u8> {
```
ali-bahjati, line 349: I think having some sort versioning here can be helpful for future changes to this message.

program/rust/src/tests/test_message.rs
guibescos: Will keep since they are also in the other tests

program/rust/src/validator.rs
guibescos: Done

program/rust/src/validator.rs
guibescos: In practice it will be a very small difference since $PYTH has 6 decimals

program/rust/src/accounts/price.rs
guibescos: Won't do it since the other messages don't have it

ali-bahjati (approved):

PR #416, opened 2024-08-05
Add feed index for batch publish

Riateche:
- Store last assigned feed index in the permission account.
- Assign new feed index when creating a new price feed.
- Add a migration instruction to assign feed index to an existing feed.

`reserve_new_price_feed_index` gets called from `init_price_feed_index` (to assign an index to an existing price) and from `add_price` (to assign an index to a new price).
The transition process would be to top up the permission account so that it's rent exempt with the new size, then to call `init_price_feed_index` for every existing price feed.

### DIFF
--- a/program/rust/src/accounts.rs
+++ b/program/rust/src/accounts.rs
@@ -105,6 +105,11 @@ pub trait PythAccount: Pod {
     /// have. `INITIAL_SIZE` <= `minimum_size()`
     const MINIMUM_SIZE: usize = size_of::<Self>();
 
+    /// Size of the account data on creation. Usually this is the same as `MINIMUM_SIZE` but it's
+    /// different for `PermissionAccount` because we've added new fields to it. In this case
+    /// we cannot increase `MINIMUM_SIZE` because that would break reading the account.
+    const NEW_ACCOUNT_SPACE: usize = Self::MINIMUM_SIZE;
+
     /// Given an `AccountInfo`, verify it is sufficiently large and has the correct discriminator.
     fn initialize<'a>(
         account: &'a AccountInfo,
@@ -139,7 +144,7 @@ pub trait PythAccount: Pod {
         seeds: &[&[u8]],
         version: u32,
     ) -> Result<(), ProgramError> {
-        let target_rent = get_rent()?.minimum_balance(Self::MINIMUM_SIZE);
+        let target_rent = get_rent()?.minimum_balance(Self::NEW_ACCOUNT_SPACE);
 
         if account.data_len() == 0 {
             create(
@@ -147,7 +152,7 @@ pub trait PythAccount: Pod {
                 account,
                 system_program,
                 program_id,
-                Self::MINIMUM_SIZE,
+                Self::NEW_ACCOUNT_SPACE,
                 target_rent,
                 seeds,
             )?;
--- a/program/rust/src/accounts.rs
+++ b/program/rust/src/accounts.rs
@@ -59,6 +59,7 @@ pub use {
         PriceEma,
         PriceInfo,
         PythOracleSerialize,
+        MAX_FEED_INDEX,
     },
     product::{
         update_product_metadata,
--- a/program/rust/src/accounts/permission.rs
+++ b/program/rust/src/accounts/permission.rs
@@ -78,7 +78,6 @@ impl PermissionAccount {
 
 impl PythAccount for PermissionAccount {
     const ACCOUNT_TYPE: u32 = PC_ACCTYPE_PERMISSIONS;
-    // TODO: change?
-    // TODO: add feed_index when creating account
-    const INITIAL_SIZE: u32 = size_of::<PermissionAccount>() as u32;
+    const INITIAL_SIZE: u32 = Self::MIN_SIZE_WITH_LAST_FEED_INDEX as u32;
+    const NEW_ACCOUNT_SPACE: usize = Self::MIN_SIZE_WITH_LAST_FEED_INDEX;
 }
--- a/program/rust/src/accounts/permission.rs
+++ b/program/rust/src/accounts/permission.rs
@@ -47,9 +47,6 @@ pub struct PermissionAccount {
 }
 
 impl PermissionAccount {
-    pub const MIN_SIZE_WITH_LAST_FEED_INDEX: usize =
-        size_of::<PermissionAccount>() + size_of::<u32>();
-
     pub fn is_authorized(&self, key: &Pubkey, command: OracleCommand) -> bool {
         #[allow(clippy::match_like_matches_macro)]
         match (*key, command) {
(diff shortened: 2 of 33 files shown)


ali-bahjati (commented):
Looks great! I'll approve once we have some tests for it too.

program/rust/src/processor/set_min_pub.rs
```diff
@@ -51,7 +51,6 @@ pub fn set_min_pub(
         &cmd.header,
     )?;
 
-
     let mut price_account_data = load_checked::<PriceAccount>(price_account, cmd.header.version)?;
```
guibescos, line 54: slightly annoyed at the formatting changes

program/rust/src/processor/init_price_feed_index.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    check_valid_funding_account(funding_account)?;
+    check_permissioned_funding_account(
+        program_id,
+        price_account,
+        funding_account,
+        permissions_account,
+        cmd,
+    )?;
+    check_valid_writable_account(program_id, permissions_account)?;
```
guibescos, line 56: currently I think `check_permissioned_funding_account` already does this check. Thought that's strange.

program/rust/src/processor.rs
```diff
@@ -84,5 +103,29 @@ pub fn process_instruction(
...
+    if permissions_account.data_len() < PermissionAccount::MIN_SIZE_WITH_LAST_FEED_INDEX {
+        let new_size = PermissionAccount::MIN_SIZE_WITH_LAST_FEED_INDEX;
+        let rent = Rent::get()?;
+        let new_minimum_balance = rent.minimum_balance(new_size);
+        pyth_assert(
+            permissions_account.lamports() >= new_minimum_balance,
+            ProgramError::AccountNotRentExempt,
+        )?;
+
+        permissions_account.realloc(new_size, true)?;
```
guibescos, line 120: nit: I think this could be false

program/rust/src/processor.rs
```diff
@@ -84,5 +103,29 @@ pub fn process_instruction(
...
+        let mut header = load_account_as_mut::<AccountHeader>(permissions_account)?;
+        header.size = try_convert(new_size)?;
     }
+    let mut last_feed_index = PermissionAccount::load_last_feed_index_mut(permissions_account)?;
+    *last_feed_index += 1;
+    pyth_assert(
+        *last_feed_index < (1 << 28),
```
guibescos, line 127: nit: magic number 28

program/rust/src/processor/add_price.rs
```diff
@@ -68,16 +70,19 @@ pub fn add_price(
         permissions_account,
         &cmd_args.header,
     )?;
+    check_valid_writable_account(program_id, permissions_account)?;
 
     let mut product_data =
         load_checked::<ProductAccount>(product_account, cmd_args.header.version)?;
 
+    // TODO: where is it created???
```
guibescos, line 78: It's created in a different instruction I think

program/rust/src/accounts/permission.rs
```diff
@@ -50,9 +60,24 @@ impl PermissionAccount {
             _ => false,
         }
     }
+
+    pub fn load_last_feed_index_mut<'a>(
```
guibescos, line 64: where is this used?

program/rust/src/processor.rs
```diff
@@ -84,5 +103,29 @@ pub fn process_instruction(
         DelProduct => del_product(program_id, accounts, instruction_data),
         UpdPermissions => upd_permissions(program_id, accounts, instruction_data),
         SetMaxLatency => set_max_latency(program_id, accounts, instruction_data),
+        InitPriceFeedIndex => init_price_feed_index(program_id, accounts, instruction_data),
+    }
+}
+
+fn reserve_new_price_feed_index(permissions_account: &AccountInfo) -> Result<u32, ProgramError> {
```
guibescos, line 110: this feels like it should be in `permission.rs`

program/rust/src/processor/add_price.rs
guibescos: Like the creator first needs to call `SystemProgram::create_account`

guibescos (commented):
I think `MIN_SIZE_WITH_LAST_FEED_INDEX`, `INITIAL_SIZE` and `NEW_ACCOUNT_SPACE` are all the same, can't we use `INITIAL_SIZE` everywhere?

program/rust/src/processor/init_price_feed_index.rs
Riateche: No, as far as I can see, `check_permissioned_funding_account` doesn't require the permission account to be writable, so a separate check is needed here.

program/rust/src/processor.rs
Riateche: I think it's safer to leave `zero_init = true` because we're not really concerned about performance here.

program/rust/src/accounts.rs
```diff
@@ -105,6 +105,11 @@ pub trait PythAccount: Pod {
     /// have. `INITIAL_SIZE` <= `minimum_size()`
     const MINIMUM_SIZE: usize = size_of::<Self>();
 
+    /// Size of the account data on creation. Usually this is the same as `MINIMUM_SIZE` but it's
+    /// different for `PermissionAccount` because we've added new fields to it. In this case
+    /// we cannot increase `MINIMUM_SIZE` because that would break reading the account.
+    const NEW_ACCOUNT_SPACE: usize = Self::MINIMUM_SIZE;
```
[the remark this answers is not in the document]
Riateche: If we increase `MINIMUM_SIZE`, that will break all instructions that read from permission account because the actual size of the permission account will be smaller than `MINIMUM_SIZE`. So that would require us to call an instruction that grows the permission account immediately after the upgrade, and I'm trying to avoid that. We can change that if it's easier to do with a migration instruction.

program/rust/src/processor.rs
Riateche: `permission.rs` is a part of `accounts` module that doesn't really contain a lot of other code like this. I think it better fits in the `processor` module.

Riateche:
> I think MIN_SIZE_WITH_LAST_FEED_INDEX, INITIAL_SIZE and NEW_ACCOUNT_SPACE are all the same, can't we use INITIAL_SIZE everywhere?

Originally, I tried to alter `INITIAL_SIZE` for `PermissionAccount`, but I realized that unfortunately `INITIAL_SIZE` is not the initial size of the account, but the initial size of the initialized data. The naming is a bit counter-intuitive. `INITIAL_SIZE` can in fact be smaller than `MINIMUM_SIZE` for accounts that have variable data length. So originally `MINIMUM_SIZE` was used both as the space to create the account and as the required minimum size when reading the account. Now that I want to support reading permission accounts with old (smaller) size but at the same time I want to create new permission accounts with new (larger) size, I have to introduce a separate constant (`NEW_ACCOUNT_SPACE`) to distinguish between these use cases specifically for the permission account. 

I've removed `MIN_SIZE_WITH_LAST_FEED_INDEX` because it's really the same thing as `NEW_ACCOUNT_SPACE` now.

program/rust/src/processor/set_min_pub.rs
Riateche: Not sure how that happened, must be rustfmt. Fixed.

ali-bahjati (approved):
Nice!

PR #429, opened 2025-09-05
feat: support zero ci

ali-behjati:
This PR adds the support for zero CI on the feeds that specifically allow it. There are some feeds that reflect a single source of truth and having a non-zero CI for them does not make sense.

The first commit adds support for Pythtest conformance that has no accumulator enabled by default so we don't need to branch off another version for it.

### DIFF
--- a/program/rust/src/tests/test_full_publisher_set.rs
+++ b/program/rust/src/tests/test_full_publisher_set.rs
@@ -79,8 +79,13 @@ async fn test_full_publisher_set() -> Result<(), Box<dyn std::error::Error>> {
             .await
             .unwrap();
 
-        assert_eq!(price_data.agg_.price_, 0);
-        assert_eq!(price_data.agg_.conf_, 0);
+        if cfg!(feature = "no-default-accumulator-v2") {
+            assert_eq!(price_data.agg_.price_, 110);
+            assert_eq!(price_data.agg_.conf_, 20);
+        } else {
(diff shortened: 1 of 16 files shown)


program/rust/src/tests/test_full_publisher_set.rs
```diff
@@ -79,8 +79,13 @@ async fn test_full_publisher_set() -> Result<(), Box<dyn std::error::Error>> {
...
 
-        assert_eq!(price_data.agg_.price_, 0);
-        assert_eq!(price_data.agg_.conf_, 0);
+        if cfg!(feature = "no-default-accumulator-v2") {
+            assert_eq!(price_data.agg_.price_, 110);
+            assert_eq!(price_data.agg_.conf_, 20);
+        } else {
+            assert_eq!(price_data.agg_.price_, 0);
+            assert_eq!(price_data.agg_.conf_, 0);
+        }
```
keyvankhademi, line 88: do we run tests with 2 different flags? One of them might not be tested.

keyvankhademi (dismissed):

Riateche (dismissed):

jayantk (dismissed):

program/rust/src/tests/test_full_publisher_set.rs
ali-behjati: yes. we test both.

Riateche (dismissed):

bplatak (approved):

